PR #1, opened 2022-09-08
boostraping tempo operator project

rubenvp8510:


frzifus (commented):
Maybe next time we should do the scaffolding in smaller step. Its quite intense to read those 2500 lines in one commit ^^

Seems we have pushed some binaries to the repo too.
image

Otherwise it looks like a good thing to start with - up to the others @pavolloffay @kevinearls

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
frzifus, line 4: Should this be under Apache License? Iam asking since the repo itself is initialized with AGPL.

https://github.com/os-observability/tempo-operator/blob/b45ee30f4107891290c0881960b819cc4a7f2c8f/LICENSE#L1-L6

Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
+FROM golang:1.18 as builder
```
iblancasa, line 2: Should we use `1.19`?

iblancasa (changes requested):
Please, add a `.gitignore` to avoid commiting the `bin` folder

rubenvp8510:
> Maybe next time we should do the scaffolding in smaller step. Its quite intense to read those 2500 lines in one commit ^^
> 
> Seems we have pushed some binaries to the repo too. image
> 
> Otherwise it looks like a good thing to start with - up to the others @pavolloffay @kevinearls

Ohh I don't think we need to commit those bins,  on the otherhand the operator sdk tool did the scaffolding, so I don't think we can do it smaller :D

Dockerfile
rubenvp8510: not sure, I mean is the most recent version of go, but I think the sdk works with 1.18

iblancasa (approved):

frzifus:
Now we need to make sure we squash it. otherwise its in the history ^^

Should we continue?

api/v1alpha1/groupversion_info.go
frzifus: Ahh wait, whats the state here?

Dockerfile
rubenvp8510: make sense, bumbed the version.

rubenvp8510:
@frzifus @iblancasa @kevinearls one last review please? If you are OK we can merge.

go.mod
```diff
@@ -0,0 +1,83 @@
+module github.com/os-observability/tempo-operator
+
+go 1.18
```
frzifus, line 3:
```suggestion
go 1.19
```

frzifus: lets bump it afterwards

frzifus (approved):

PR #16, opened 2022-09-13
Support crdoc

### LINKED ISSUE #7: Create API/CRD docs

frzifus:
Closes #7 

preview: https://github.com/os-observability/tempo-operator/blob/d4a3c78e9c644c857dc1bc77048e9bbd72d32d85/docs/api.md


frzifus:
oh missing operator-sdk :X

Makefile
```diff
@@ -237,3 +238,47 @@ catalog-push: ## Push a catalog image.
...
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-get-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+go get -d $(2)@$(3) ;\
+GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
```
iblancasa, line 272: The `LOCALBIN` variable is there. You don't need to define `PROJECT_DIR`.

Makefile
```diff
@@ -237,3 +238,47 @@ catalog-push: ## Push a catalog image.
...
+	$(call go-get-tool,$(CRDOC), fybrik.io/crdoc,$(CRDOC_VERSION))
+
+
+.PHONY: kustomize
+kustomize: ## Download kustomize locally if necessary.
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4,$(KUSTOMIZE_VERSION))
+
+# go-get-tool will 'go get' any package $2 and install it to $1.
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-get-tool
```
iblancasa, line 264: I know other projects create this macro to download tools but... could we declare it in a separate shell script? That will allow us to add retry logic.
Or, probably, we can use [bingo](https://github.com/bwplotka/bingo) as Observatorium does

Makefile
```diff
@@ -237,3 +238,47 @@ catalog-push: ## Push a catalog image.
...
+rm -rf $$TMP_DIR ;\
+}
+endef
+
+.PHONY: ensure-generate-is-noop
+ensure-generate-is-noop: generate bundle
+	@# on make bundle config/manager/kustomization.yaml includes changes, which should be ignored for the below check
+	@git restore config/manager/kustomization.yaml
+	@git diff -s --exit-code api/v1alpha1/zz_generated.*.go || (echo "Build failed: a model has been changed but the generated resources aren't up to date. Run 'make generate' and update your PR." && exit 1)
+	@git diff -s --exit-code bundle config || (echo "Build failed: the bundle, config files has been changed but the generated bundle, config files aren't up to date. Run 'make bundle' and update your PR." && git diff && exit 1)
```
iblancasa, line 282: Just asking, because I know for other operators is needed: does the `USER` environment variable needs an special value for calling `make bundle` and make this rule not fail?

Makefile
```diff
@@ -102,7 +102,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests.
+test: manifests generate fmt vet ensure-generate-is-noop envtest ## Run tests.
```
iblancasa, line 105: Why `ensure-generate-is-noop` is needed for running the unit tests?

Makefile
```diff
@@ -237,3 +238,47 @@ catalog-push: ## Push a catalog image.
...
+
+
+CRDOC = $(shell pwd)/bin/crdoc
+.PHONY: crdoc
+crdoc: ## Download crdoc locally if necessary.
+	$(call go-get-tool,$(CRDOC), fybrik.io/crdoc,$(CRDOC_VERSION))
+
+
+.PHONY: kustomize
+kustomize: ## Download kustomize locally if necessary.
```
iblancasa, line 259: Could you add the logic to not download the binary if not needed?
[Reference](https://v1-23-x.sdk.operatorframework.io/docs/upgrading-sdk-version/v1.23.0/#gov3-fix-makefile-targets-to-not-download-binaries-when-they-are-found-already)

iblancasa (changes requested):

Makefile
frzifus: Can we not add some retry logic to `go-get-tool` if its needed?

Makefile
frzifus: good point, in other repos its used for the image name.

Makefile
frzifus: Why not? that we you just call `make test` and get a result if your state is ready to commit.

Makefile
iblancasa: I know. Is something we need to take into account? Probably to add it to this message or something

Makefile
iblancasa: From the experience we have with GitHub Actions, it is really possible we will need it.

Makefile
iblancasa: Because if you are doing quick changes to test stuff, you'll need to commit the results all the time in order to really run the rules from `make test`.
For me, it would make more sense to create a git hook instead of forcing the `ensure-generate-is-noop` rule here.

Makefile
frzifus: i will quickly add a retry logic.

Makefile
frzifus: done

Makefile
kevinearls: I agree with @iblancasa here.  Having to commit changes just to run the unit-tests is a nuisance.

Makefile
kevinearls: It would be great if for tools we could be great if we could check if the right version of a tool is installed and if not then go get it.  Maybe the bingo too that @iblancasa mentioned above could be used for this.

Makefile
frzifus: I followed here the [OpenTelemetry-Operator](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.59.0/Makefile#L86) approach. According to my understanding the sense behind it is that `make test` calls all needed tests. To test parts, you can then fall back on e.g. manual cli commands. But should be a big deal to register it as an extra ci step.

Git hooks would prevent you from creating an extra generation commit. Which makes the review quite painful.

iblancasa: If the problem is related to the CI, just add this dependency to the `ci` Make target. 

> Git hooks would prevent you from creating an extra generation commit. Which makes the review quite painful.

I don't think so. It will prevent you of making a commit of non checked in files.
From what I understand, you don't like this approach because you will have one commit for your changes + another one for committing the offending files for `ensure-generate-is-noop`, right? But if you force to commit them each time we want to run `make test`, we will have a commit for each change done affecting the tests and the files generation.

I think the correct solution should be add this dependency to the `ci` target.

Makefile
iblancasa: Why is a problem to add it now?

Makefile
frzifus: seems it doesnt matter, since the its not part of the `${IMG}` variable.

Makefile
frzifus: Until a few minutes ago, that was my plan. However, I have the feeling that this pr is getting bigger and bigger. So it makes sense from my point of view to deliver this afterwards.

frzifus:
mh...
```bash
STEP: bootstrapping test environment
1.6630855143536432e+09	DEBUG	controller-runtime.test-env	starting control plane
1.6630855143549504e+09	ERROR	controller-runtime.test-env	unable to start the controlplane	{"tries": 0, "error": "exec: \"etcd\": executable file not found in $PATH"}
```

Makefile
```diff
@@ -153,37 +153,33 @@ LOCALBIN ?= $(shell pwd)/bin
...
+controller-gen: ## Download controller-gen locally if necessary.
+	test -s $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION) || $(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen,$(CONTROLLER_TOOLS_VERSION))
 
-.PHONY: envtest
-envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
-$(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+.PHONY: setup-envtest
+setup-envtest: ## Download envtest-setup locally if necessary.
+	test -s $(LOCALBIN)/setup-envtest-latest || $(call go-get-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,latest)
```
rubenvp8510, line 175: When I run this on my computer it installs `bin/setup-envtest-latest` but then in your CI target you use
`bin/setup-envtest-1.24.2`

rubenvp8510: `ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_K8S_VERSION)`

rubenvp8510:
What do you think about having the dependencies installation in a couple of scripts into hack?  I think @iblancasa did something similar for jaeger operatror https://github.com/jaegertracing/jaeger-operator/tree/main/hack/install

If not our Makefile will ends with a lot of install dependencies scripts and will be a huge Makefile.

Makefile
pavolloffay: Does bingo support retries?

pavolloffay:
The PR needs to be rebased and I would suggest to split it into separate PRs (generate CR docs, change how tools are installed, fix bundle manifests)

Makefile
iblancasa: @frzifus how about creating a new task to not forget about this?
@pavolloffay I'm not sure but I think it has some kind of "recovery logic". It seems bingo is becoming something more standard. But using bingo was just a suggestion about something that could be interesting to implement. No strong opinion

iblancasa:
@rubenvp8510 I agree. That was what I was suggesting in https://github.com/os-observability/tempo-operator/pull/16#discussion_r969429077. It is something that doesn't need to be done in this PR. As @pavolloffay points, it can be a good idea to split that work in different PRs :)

Makefile
pavolloffay: I am not a huge fan of bingo, at least I don't see the benefits of introducing another tooling. The go package version can be pinned directly in go get.

frzifus: I agree, its not to complex to download and install specific versions. i would argue with a retry logic. At the end a for loop with a timeout is a simple and sufficient solution.

frzifus:
See: https://github.com/os-observability/tempo-operator/pull/23 https://github.com/os-observability/tempo-operator/pull/24 https://github.com/os-observability/tempo-operator/pull/28

PR #18, opened 2022-09-13
Avoid fast cache growing in GitHub Actions Workflow Cache

### LINKED ISSUE #17: Avoid fast cache growing in publish-images GitHub Actions Workflow
While the implementation of #15, the Docker cache was used to speedup the builds. [There is an issue](https://github.com/docker/build-push-action/issues/252) associated to the `actions/cache@v3` GitHub Action that makes the cache grown pretty fast.

To avoid this issue, we have to add this hack to the end of the GitHub Actions Workflow

iblancasa:

Resolves #17

### DIFF
--- a/.github/workflows/publish-images.yaml
+++ b/.github/workflows/publish-images.yaml
@@ -68,3 +68,12 @@ jobs:
           labels: ${{ steps.docker_meta.outputs.labels }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+      -
+        # Temp fix
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-images.yaml
```diff
@@ -68,3 +68,12 @@ jobs:
           labels: ${{ steps.docker_meta.outputs.labels }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
+      -
+        # Temp fix
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
+        name: Move cache
```
pavolloffay, line 75: I didn't see this being a problem for us, but we can be proactive and add it here.

pavolloffay (approved):

PR #45, opened 2022-09-23
makefile: fix ignore-not-found issue

frzifus:


Makefile
```diff
@@ -125,9 +125,7 @@ docker-push: ## Push docker image with the manager.
 
 ##@ Deployment
 
-ifndef ignore-not-found
```
pavolloffay, line 128: Other projects use this form without causing issues. Most likely there is a different root cause that is causing this. I would prefer to find the root case and fix that.

```
ignore-not-found = false
bash: line 1: ignore-not-found: command not found
make: *** [Makefile:125: docker-push] Error 127
```

frzifus: Seems to be a tab/space issue.

pavolloffay (approved):

PR #46, opened 2022-09-26
WIP: Add microservices reconcile controller

pavolloffay:
This minimal implementation handles only the ingestion part - distributor and ingester.


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: docker.io/pavolloffay/tempo-operator
+  newTag: dev-3
```
frzifus, line 16: should we change that back?

internal/manifests/ingester/ingester.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/os-observability/tempo-operator/api/v1alpha1"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func BuildIngester(tempo v1alpha1.Microservices) []client.Object {
```
frzifus, line 15: Should we add some comments to exposed methods? it might be obvious sometimes. But when people jump into new projects, its nice to have a bit more context.

frzifus (commented):
Had just a brief look.

controllers/microservices_controller.go
```diff
@@ -31,16 +39,72 @@ type MicroservicesReconciler struct {
...
+
+	objects, err := manifests.BuildAll(manifests.Params{Tempo: tempo})
+	// TODO(pavolloffay) check error type and change return appropriately
+	if err != nil {
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: time.Second,
+		}, err
+	}
+	errCount := 0
```
frzifus, line 64: Is `errCount` used? Looks like its increased in case of failure. But never used?

controllers/microservices_controller.go
```diff
@@ -31,16 +39,72 @@ type MicroservicesReconciler struct {
...
 func (r *MicroservicesReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&tempov1alpha1.Microservices{}).
+		For(&v1alpha1.Microservices{}).
 		Complete(r)
 }
+
+func isNamespaceScoped(obj client.Object) bool {
+	switch obj.(type) {
+	case *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding:
```
frzifus, line 105: Isnt there another way to figure out what objects are namespace scoped? Its just a feeling, but `isNamespaceScoped` does only consider 2 types and feels misleading.

internal/manifests/config/configmap.go
```diff
@@ -0,0 +1,214 @@
...
+	kvStore = "memberlist"
+)
+
+// nolint
+func config(tempo v1alpha1.Microservices) (string, error) {
+	logLevel := logging.Level{}
+	logLevel.Set("warn")
+	logFormat := logging.Format{}
+	logFormat.Set("logfmt")
+	cfg := tempoapp.Config{
```
frzifus, line 41: Will the `tempoApp.Config` be backward compatible in the long run? Otherwise it could lead to problems when older component images are used.

internal/manifests/mutate.go
```diff
@@ -0,0 +1,226 @@
...
+		switch existing.(type) {
+		case *corev1.ConfigMap:
+			cm := existing.(*corev1.ConfigMap)
+			wantCm := desired.(*corev1.ConfigMap)
+			mutateConfigMap(cm, wantCm)
+
+		case *corev1.Service:
+			svc := existing.(*corev1.Service)
+			wantSvc := desired.(*corev1.Service)
+			return mutateService(svc, wantSvc)
```
frzifus, line 53: Just a question, why is `mutateService` returned and all the other mutate functions not?

frzifus (commented):
Overall lgtm, just a few questions :)

controllers/microservices_controller.go
pavolloffay: good catch, the count should be checked

controllers/microservices_controller.go
pavolloffay: Do you mean to figure out the scope from the k8s API? 

I don't think there is a way.

internal/manifests/config/configmap.go
pavolloffay: Good comment. I don't know, probably the new config options are ignored when parsing the config. There will be a separate follow up task to improve config handling.

internal/manifests/mutate.go
pavolloffay: The other functions don't return error.

internal/manifests/config/configmap.go
frzifus: new options are not an issue. I thought more about removed and renamed options ^^

controllers/microservices_controller.go
frzifus: did you add a check?

internal/manifests/config/configmap.go
pavolloffay: Also we want to probably support an operator version per tempo version.

controllers/microservices_controller.go
pavolloffay: yes

pavolloffay: Pushed the change now

frzifus (approved):

PR #57, opened 2022-10-02
Use kubebuilder component config

rubenvp8510:


main.go
```diff
@@ -33,41 +33,32 @@ func init() {
 }
 
 func main() {
-	var metricsAddr string
-	var enableLeaderElection bool
-	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
pavolloffay, line 39: These config options should be moved to the config file and flags could be kept here to override the config from the file.

api/v1alpha1/tempooperatorconfig_types.go
```diff
@@ -0,0 +1,39 @@
+package v1alpha1
```
pavolloffay, line 1: I would prefer to have this in a different package e.g. /api/config/...

rubenvp8510: mmh not sure, this is a Kind, as the other one, wondering the tools for generate manifests etc.. will break with this change.

pavolloffay: we might need to change the api path, but it should be possible to put it in any directory.

rubenvp8510: it should be possible, but it will break a little bit the directory structure, `api/<version>/ ` also, this is part of the API, is another CRD that belongs to this operator.

main.go
rubenvp8510: Done, I reintroduced the flags.

api/v1alpha1/tempooperatorconfig_types.go
pavolloffay: It is part of the operator API, not the tempo API, that is the motivation to move it to a separate package to isolate them.

PROJECT
```diff
@@ -20,4 +20,12 @@ resources:
     defaulting: true
     validation: true
     webhookVersion: v1
+- api:
```
pavolloffay, line 23: maybe we are missing some config here? 

https://github.com/grafana/loki/blob/main/operator/PROJECT#L1

pavolloffay: https://book.kubebuilder.io/reference/project-config.html

PROJECT
```diff
@@ -20,4 +20,12 @@ resources:
...
+    crdVersion: v1
+    namespaced: true
+  domain: grafana.com
+  group: tempo
+  kind: TempoOperatorConfig
```
pavolloffay, line 28: we should remove Tempo prefix

api/v1alpha1/tempooperatorconfig_types.go
frzifus: Operator-sdk has a option to define a group. Iam not 100% sure, but would it be a different group?

```
./bin/operator-sdk-v??? create api --group config --version v1alpha1 --kind OperatorConfig
```

main.go
```diff
@@ -29,45 +30,50 @@ func init() {
...
 	utilruntime.Must(tempov1alpha1.AddToScheme(scheme))
+	utilruntime.Must(configtempov1alpha1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
 }
 
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var configFile string
```
pavolloffay, line 41: The values from flags should be probably propagated to the config file.

e.g. the config file has an option to set metrics, enable leader election, health probe.

kevinearls (approved):
LGTM, but for a PR with this many changes it would be good to have a longer description or maybe better an issue describing what is going on here.

main.go
rubenvp8510: What do you mean? If the config file specify those values it will override the flags, otherwise the flag values will be the ones that will be used.

apis/config.tempo/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,20 @@
+// Package v1alpha1 contains API Schema definitions for the config.tempo v1alpha1 API group.
+// +kubebuilder:object:generate=true
+// +groupName=config.tempo.grafana.com
+package v1alpha1
```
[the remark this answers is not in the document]
rubenvp8510: is because the group is config.tempo. I can rename the folder if this is a problem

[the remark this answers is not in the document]
rubenvp8510: Done.

pavolloffay (approved):

frzifus:
Why did we rename the `api` folder to `apis`?

rubenvp8510:
Is a convention on multigroup projects https://book.kubebuilder.io/migration/multi-group.html

if the project is not new and has previous APIs already implemented will be in the previous structure. Notice that with the multi-group project the Kind API’s files are created under apis/<group>/<version> instead of api/<version>. Also, note that the controllers will be created under controllers/<group> instead of controllers.

PR #91, opened 2022-10-17
Add initial e2e tests for default deployment and one with Jaeger

kevinearls:
Signed-off-by: Kevin Earls <kearls@redhat.com>


tests/e2e/smoketest-with-jaeger/03-run-check-services.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # This is a hacky but simple way of determining if any of the traces created by tracegen were received.  It should be considered temporary.
+  - command: 'kubectl exec --namespace $NAMESPACE deployment/tempo-simplest-query-frontend --container tempo-query -- sh  -c "rm -f services ; wget -q http://localhost:16686/api/services ; cat services | grep -i tracegen"'
```
iblancasa, line 5: It would be better if you get the route from the Ingress. Otherwise, it'll not work with a Kubernetes cluster run in a different machine than the one used to run the tests

tests/e2e/smoketest-simple/00-install-minio.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: "kubectl create --namespace $NAMESPACE -f ../minio.yaml"
```
iblancasa, line 4: Just a question: would it make sense to move this logic to a `make` rule? I mean: if this is used just for the E2E tests, is fine. But if we expect people to run their "hello world" deploying `minio` first, probably it makes sense to add a `make` rule to make everything easier

iblancasa (changes requested):

tests/e2e/smoketest-with-jaeger/03-run-check-services.yaml
kevinearls: Yes, but since I'm doing a "kubectl exec" it's going to be executed in the same container, right?  And as I noted, this is just a hack for until we come up with a better solution, which I want to talk to you about shortly.

tests/e2e/smoketest-simple/00-install-minio.yaml
kevinearls: Maybe?  I'm not really sure what will be done longterm.  If we end up adding a Makefile just for tests it could certainly go there.  I also need to look into whether you can deploy a single minio instance that can be used cluster wide.

tests/e2e/smoketest-with-jaeger/03-run-check-services.yaml
iblancasa: Oh! Good point! Sorry. I didn't notice it was done using `exec`

frzifus: nit, but would that work too? Seems to be simpler ^^
```suggestion
  - command: 'kubectl exec --namespace $NAMESPACE deployment/tempo-simplest-query-frontend --container tempo-query -- sh  -c "curl http://localhost:16686/api/services | grep -i tracegen"'
```

tests/e2e/smoketest-simple/00-install-minio.yaml
frzifus: It also would give us some flexibility on the storage side. We could use something like `../minio_losing_blocks.yaml` while testing for the compactor. (but maybe thats out of scope)

tests/e2e/smoketest-with-jaeger/03-run-check-services.yaml
kevinearls: @frzifus I don't think curl exists on those images, and whatever version of wget the have does not support -O

tests/e2e/smoketest-simple/00-install-minio.yaml
kevinearls: @frzifus  after chatting with @iblancasa I've added a make target and am leaving it so that minio is installed in the namespace for each test.  The minio install is quite fast so it shouldn't be a problem.

kevinearls:
@pavolloffay @rubenvp8510 @frzifus Please review and feel free to leave suggestions for additional tests

Makefile
```diff
@@ -187,6 +187,9 @@ KUTTL ?= $(LOCALBIN)/kubectl-kuttl
 export KUBE_VERSION ?= 1.25
 KIND_CONFIG ?= kind-$(KUBE_VERSION).yaml
 
+# Default namespace for minio install
+MINIO_NAMESPACE ?= default
```
pavolloffay, line 191: We could change this and install minio in a different namespace. The default namespace does not work on OCP. The minio manifests should create namespace where the instance will be created.

tests/e2e/smoketest-simple/01-assert.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: tempo.grafana.com/v1alpha1
+kind: Microservices
+metadata:
+  name: simplest
+---
```
pavolloffay, line 5: Could you please assert:
* services
* exposed ports
* labels on the object meta
* match labels

pavolloffay (changes requested):
I would suggest removing one test and keep only the one with Jaeger-query.

There should be other unit tests that verify jaeger query gets correctly enabled/disabled based on the CR.

Makefile
iblancasa: I agree. Probably the `storage` namespace is a good one?

pavolloffay: or just `minio` to be idiomatic.

kevinearls:
@pavolloffay Can you take another look when you have a chance?

pavolloffay (approved):
LGTM

Please open a follow-up issue to improve test time. The tracegen runs now 30s which seems way too long. Better would be to implement retry in the assert.

PR #95, opened 2022-10-20
support global block retention

frzifus:
Signed-off-by: Benedikt Bongartz <bongartz@klimlive.de>

### DIFF
--- a/apis/tempo/v1alpha1/microservices_types.go
+++ b/apis/tempo/v1alpha1/microservices_types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -306,7 +308,7 @@ type RetentionConfig struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:text",displayName="Trace Retention Period"
-	Traces string `json:"traces"`
+	Traces time.Duration `json:"traces"`
 }
 
 func init() {
--- a/apis/tempo/v1alpha1/microservices_types.go
+++ b/apis/tempo/v1alpha1/microservices_types.go
@@ -304,6 +304,7 @@ type RetentionSpec struct {
 type RetentionConfig struct {
 	// Traces defines retention period. Supported parameter suffixes are “s”, “m” and “h”.
 	// example: 336h
+	// default: value is 48h.
 	//
 	// +optional
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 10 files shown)


internal/manifests/config/configmap.go
```diff
@@ -160,9 +160,15 @@ func BuildConfigs(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMa
 		s3Endpoint = strings.TrimPrefix(s3Endpoint, "https://")
 	}
 
+	globalRetention := tempo.Spec.Retention.Global.Traces
+	if globalRetention == "" {
```
pavolloffay, line 164: This should be done in the defaulting webhook and the type of `tempo.Spec.Retention.Global.Traces` should be probably golang duration and not string to enable validation.

pavolloffay (changes requested):

internal/manifests/config/configmap.go
frzifus: fair point :+1:

frzifus: changed the trace type from `string` to `time.Duration`. Its a good thing, since they use `time.Duration` also in the tempo config.

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -25,8 +27,9 @@ var _ webhook.Defaulter = &Microservices{}
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *Microservices) Default() {
 	microserviceslog.Info("default", "name", r.Name)
-
-	// TODO(user): fill in your defaulting logic.
+	if r.Spec.Retention.Global.Traces == 0 {
```
pavolloffay, line 30: can be retention set to 0? In some systems it means unlimited retention. If that is the case we should change `Traces` to a pointer.

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -306,7 +308,7 @@ type RetentionConfig struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:text",displayName="Trace Retention Period"
-	Traces string `json:"traces"`
+	Traces time.Duration `json:"traces"`
```
pavolloffay, line 311: Add  default to the godoc.

apis/tempo/v1alpha1/microservices_webhook.go
frzifus: They dont consider 0 at all.
https://github.com/grafana/tempo/blob/9c296831384c29dc9b706601b1580b83b77b9d90/modules/compactor/config.go#L36
https://github.com/grafana/tempo/blob/9c296831384c29dc9b706601b1580b83b77b9d90/pkg/util/config.go#L70-L76

pavolloffay (approved):

PR #106, opened 2022-10-27
Define tempo config using a template

rubenvp8510:


internal/manifests/config/build.go
```diff
@@ -0,0 +1,36 @@
+package config
+
+import (
+	"bytes"
+	"embed"
+	"html/template"
```
frzifus, line 6: whats the difference to `text/template` ?

frzifus (approved):
like it :)

internal/manifests/config/configmap.go
```diff
@@ -2,172 +2,43 @@ package config
...
-func BuildConfigs(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMap, error) {
-	s3Insecure := false
-	s3Endpoint := params.S3.Endpoint
-	if strings.HasPrefix(s3Endpoint, "http://") {
-		s3Insecure = true
-		s3Endpoint = strings.TrimPrefix(s3Endpoint, "http://")
-	} else if !strings.HasPrefix(s3Endpoint, "https://") {
-		s3Insecure = true
+// BuildConfigMap creates configuration objects.
+func BuildConfigMap(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMap, error) {
```
pavolloffay, line 20: It seems that we are leaking some APIs or abstractions here. 

```go
// Params holds configuration parameters.
type Params struct {
	S3 S3Options
}

// S3 holds S3 object storage configuration options.
type S3Options struct {
	Endpoint string
	Bucket   string
	Insecure bool
}

// Options holds the configuration template options.
type Options struct {
	GlobalRetention        string
	QueryFrontendDiscovery string
	MemberList             []string
	S3                     S3Options
}
```

I don't like that `Params` encapsulate `S3Options` that are used in `Options` for the rendering. Could be `Options` defined as private (with all it's dependencies)?

rubenvp8510: Done

internal/manifests/config/build.go
rubenvp8510: Good question, I think the main difference is that html template will scape some characters

pavolloffay (approved):

frzifus (approved):

PR #113, opened 2022-11-10
Configure search for tempo

rubenvp8510:


apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -104,42 +104,41 @@ type SearchSpec struct {
 	// +optional
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Query timeout"
-	QueryTimeout time.Duration `json:"queryTimeout,omitempty"`
+	QueryTimeout string `json:"queryTimeout,omitempty"`
```
frzifus, line 107: Do we avoid the datatype here so that in the end the configuration looks better? Means a value would be `1m5s` instead of `65`?

pavolloffay (commented):
What is the default behavior if the search is not configured in the CR? 

The operator should provide an optioned deployment (e.g. with search enabled with good enough configuration for majority of deployments)

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -128,6 +134,21 @@ type Resources struct {
 	Total *corev1.ResourceRequirements `json:"total,omitempty"`
 }
 
+type SearchSpec struct {
```
pavolloffay, line 137: please add godoc to the public members of the struct and to the struct as well

pavolloffay: Are all these settings global or per tenant?

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -128,6 +134,21 @@ type Resources struct {
...
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Max search time range allowed"
+	MaxSearchDuration string `json:"maxSearchDuration,omitempty"`
```
pavolloffay, line 141: What is the default value?

Does 0 have a special meaning? 

Is this a lookback, meaning that the range can be always from the current time or any range in the past?

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -128,6 +134,21 @@ type Resources struct {
...
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Limit used for search requests if none is set by the caller"
+	DefaultResultLimit int `json:"defaultResultLimit,omitempty"`
```
pavolloffay, line 145: What is the default value? 

What is the search request?  A request to the object store?

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -128,6 +134,21 @@ type Resources struct {
...
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="The maximum allowed value of the limit parameter on search requests"
+	MaxResultLimit int `json:"maxResultLimit,omitempty"`
```
pavolloffay, line 149: what is the default value?

pavolloffay (commented):
The duration values should be validated in the webhook.

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: The default value is 0, which means unlimited, perhaps we should set it to the retention period?

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: The default value is the default value of tempo configuration (which in this case is 20) https://grafana.com/docs/tempo/latest/configuration/#query-frontend

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: The default value is the value indicated in the tempo config docs , in this case this limit is disabled ( value = 0)

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: @rubenvp8510 take a look at this PR https://github.com/os-observability/tempo-operator/pull/145

The `time.Duration` can be properly exposed via wrapper `metav1.Duration`.

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -125,6 +131,35 @@ type Resources struct {
...
 }
 
+type SearchSpec struct {
+	// Enable tempo search feature, default to true
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Search enabled"
+	Enabled *bool `json:"enabled,omitempty"`
+	// Limit used for search requests if none is set by the caller (default: 20)
```
pavolloffay, line 141: in which cases none is set by the caller? E.g. jaeger-ui would not set it?

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: Is this related to `DefaultResultLimit` ? If so it makes sense to colocate them and crosslink them in the godoc.

pavolloffay (commented):
LGTM, just minor comments about godoc. 


@rubenvp8510 it would be good to create  follow-up issues as we have discussed during the vf2f.

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: It could be on jaeger-ui or if someone else calls directly the API.

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: Not really, this is the max number of traces you can fetch from tempo, I think the propose of this is  to artificially limit the results in order to prevent abuses, (queries that returns a lot of traces which could overload the system)

rubenvp8510:
Created an issue to track the remaining work related to search configuration, which is defined how to property set some search parameters based on amount of data/ resources assigned. https://github.com/os-observability/tempo-operator/issues/152

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: As a newbie, I am still confused about what this limit means. Does it limit the number of returned traces or spans or any other data types?

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: I have the same comment as for `DefaultResultLimit` we should mention what it actually limits - traces/spans.

apis/tempo/v1alpha1/microservices_types.go
rubenvp8510: It is on traces, already specified in the comment.

pavolloffay (approved):

PR #114, opened 2022-11-10
Update README: fix minio endpoint

andreasgerstmayr:
Minio is deployed to the 'minio' namespace when following the instructions in the README.


frzifus (approved):
good catch :)

PR #116, opened 2022-11-14
Ensure that the overrides key is a map

andreasgerstmayr:
Tempo parses the CLI arguments and sets up the config object with values from the CLI arguments or their default values. Later, this config object is overlayed with the unmarshalled yaml configuration file.

An empty overrides key in the yaml:
```
overrides:
```
sets overrides to nil instead of an empty map. `yaml.Unmarshal` unmarshals nil values to structs by setting every struct member to the zero value.

This results in overwriting the default values, for example the setting `ingestion_rate_limit_bytes` will be set to 0 instead of the default value as provided by tempo.


andreasgerstmayr:
Surrounding the map with `{` and `}` doesn't work if there are multiple map entries, for example:
```
overrides: {
  ingestion_burst_size_bytes: 100
  ingestion_rate_limit_bytes: 200
}
```
In this case commas are required between each map entry, otherwise the config is invalid YAML. Adding these commas at the right place is troublesome with Go templating (i.e. the last entry must not have a comma), therefore I've changed the code of this PR to only include the `overrides` key if at least one override is set.

frzifus (approved):
lgtm :)

Issue #100, opened 2022-10-24
Make docker images configurable


pavolloffay:
Add tempo and Jaeger query images to the:

- [ ] CR
- [ ] operator config

PR #121, opened 2022-11-23
Make container images configurable in the CR

andreasgerstmayr:
Resolves: #100

Currently we have a unit test for the default values. Do we need another unit test for the entire defaulter webhook?


apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -94,6 +100,18 @@ type MicroservicesList struct {
 	Items           []Microservices `json:"items"`
 }
 
+type ImagesSpec struct {
```
frzifus, line 103: lets put a comment on top and describe what its used for.

main.go
```diff
@@ -79,6 +79,7 @@ func main() {
 			setupLog.Error(err, "unable to create webhook", "webhook", "Microservices")
```
frzifus, line 79:
```suggestion
			setupLog.Error(err, "unable to create validator webhook", "webhook", "Microservices")
```

main.go
```diff
@@ -79,6 +79,7 @@ func main() {
...
 			os.Exit(1)
 		}
+		mgr.GetWebhookServer().Register("/mutate-tempo-grafana-com-v1alpha1-microservices", tempov1alpha1.NewMicroservicesDefaulterWebhook(mgr.GetClient(), ctrlConfig.DefaultImages))
```
frzifus, line 82: Does register return an error that we should handle?

frzifus (commented):
i like :)

andreasgerstmayr:
Just after opening the PR I realized afaics I can also use a custom defaulter:
```
return ctrl.NewWebhookManagedBy(mgr).
		For(r).
		WithDefaulter(&defaulter{defaultImages: defaultImages}).
		Complete()
```
saving us a bit of boilerplate code in handling the webhook. I'll update this PR shortly.

andreasgerstmayr:
Updated the PR with a custom defaulter, so we can keep using the webhook builder, get a generated webhook path and in general need less boilerplate.

andreasgerstmayr:
Fixed the linter error.

Thanks for the review  @frzifus!

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -11,28 +15,50 @@ import (
 )
 
 var (
-	zeroQuantity  = resource.MustParse("0Gi")
-	tenGBQuantity = resource.MustParse("10Gi")
+	zeroQuantity                = resource.MustParse("0Gi")
+	tenGBQuantity               = resource.MustParse("10Gi")
+	ErrNoDefaultTempoImage      = errors.New("please specify a tempo image in the CR or in the operator configuration")
+	ErrNoDefaultTempoQueryImage = errors.New("please specify a tempo-query image in the CR or in the operator configuration")
```
frzifus, line 21:
```suggestion
	errNoDefaultTempoImage      = errors.New("please specify a tempo image in the CR or in the operator configuration")
	errNoDefaultTempoQueryImage = errors.New("please specify a tempo-query image in the CR or in the operator configuration")
```

We do not use this errors somewhere else, right? If so, i would recommend to keep them private.

frzifus (approved):
awesome :) Ive only one tiny thing ^^

apis/tempo/v1alpha1/microservices_webhook.go
andreasgerstmayr: Good idea, I've updated the PR.

Issue #71, opened 2022-10-10
Make gossip match label instance aware


pavolloffay:
Currently, the operator uses `tempo-gossip-member: 'true'` label selector for the gossip.

If two tempo instances are created in the same namespace this will create issues, hence the label should be changed to e.g. `tempo-${instance}-gossip-member`

PR #123, opened 2022-11-25
Add common labels (including instance name) to the gossip ring selector

andreasgerstmayr:
This is required to support multiple tempo instances in the same namespace.

Instead of updating the label name, it makes the selector of the gossip-ring service more strict to only match pods created by this controller (and with the same instance name). Is there any drawback to this vs. renaming the label per-instance?

Resolves: #71

### DIFF
--- a/tests/e2e/smoketest-with-jaeger/01-assert.yaml
+++ b/tests/e2e/smoketest-with-jaeger/01-assert.yaml
@@ -161,6 +161,10 @@ spec:
       protocol: TCP
       targetPort: http-memberlist
   selector:
+    app.kubernetes.io/created-by: tempo-controller
+    app.kubernetes.io/instance: simplest
+    app.kubernetes.io/managed-by: tempo-controller
+    app.kubernetes.io/name: tempo
     tempo-gossip-member: "true"
 ---
 apiVersion: v1
--- a/internal/manifests/manifestutils/naming.go
+++ b/internal/manifests/manifestutils/naming.go
@@ -16,12 +16,12 @@ func Name(component string, instanceName string) string {
 
 // ComponentLabels is a list of all commonLabels including the app.kubernetes.io/component:<component> label.
 func ComponentLabels(component, instanceName string) labels.Set {
-	return labels.Merge(commonLabels(instanceName), map[string]string{
+	return labels.Merge(CommonLabels(instanceName), map[string]string{
 		"app.kubernetes.io/component": component,
 	})
 }
 
-func commonLabels(instanceName string) map[string]string {
+func CommonLabels(instanceName string) map[string]string {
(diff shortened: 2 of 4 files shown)


tests/e2e/smoketest-with-jaeger/01-assert.yaml
```diff
@@ -161,6 +161,10 @@ spec:
       protocol: TCP
       targetPort: http-memberlist
   selector:
+    app.kubernetes.io/created-by: tempo-controller
+    app.kubernetes.io/instance: simplest
```
frzifus, line 165: Couldnt that lead to a conflict with other applications like a jaeger instance called `simplest`? I dont know the convention for `app.kubernetes.io/instance`. ^^

andreasgerstmayr: all labels need to match, so it would only create a conflict if the jaeger instance also has all 5 labels set:
```
app.kubernetes.io/created-by: tempo-controller
app.kubernetes.io/instance: simplest
app.kubernetes.io/managed-by: tempo-controller
app.kubernetes.io/name: tempo
tempo-gossip-member: "true"
```
afaics this should never be the case?

frzifus: right, but if i would filter for instance=simplest in another application/cli tool or eq this could be the case. To make this clear, ive not strong opinion on that. ^^ Its more a stupid question. :smile: 
image
Found that thing here: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

andreasgerstmayr: I peeked at the official tempo helm chart, and they're using the following selector for the gossip service:
```
selector:
  app.kubernetes.io/instance: my-release
  app.kubernetes.io/name: tempo
  app.kubernetes.io/part-of: memberlist
```

and pods have the following labels (e.g. the distributor pod):
```
labels:
  app.kubernetes.io/component: distributor
  app.kubernetes.io/instance: my-release
  app.kubernetes.io/managed-by: Helm
  app.kubernetes.io/name: tempo
  app.kubernetes.io/part-of: memberlist
  app.kubernetes.io/version: 1.5.0
  helm.sh/chart: tempo-distributed-0.27.9
  pod-template-hash: 79c5d88cd4
```

pavolloffay (approved):

tests/e2e/smoketest-with-jaeger/01-assert.yaml
frzifus: thanks :)

frzifus (approved):

Issue #58, opened 2022-10-03
Add validation of storage secret to webhook


pavolloffay:
Storage secret should be validated in the webhook.

The webhook should check if the secret exists and the right content is provided - e.g. the endpoint is a valid endpoint etc.

frzifus:
i can work on that :+1:

PR #127, opened 2022-12-02
Validate storage secret

andreasgerstmayr:
* checks if the secret name is set and refers to an existing secret
* checks if the secret content is empty
* checks if all required fields are set in the secret
* checks if the endpoint is a valid URL

The branch of this PR is based on #124 because it also requires the custom validator.
Resolves: #58

### DIFF
--- a/apis/tempo/v1alpha1/microservices_types.go
+++ b/apis/tempo/v1alpha1/microservices_types.go
@@ -142,7 +142,7 @@ type ObjectStorageSpec struct {
 	// Name of a secret in the same namespace as the tempo Microservices custom resource.
 	//
 	// +optional
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Object Storage Secret"
 	Secret string `json:"secret,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


frzifus:
@andreasgerstmayr could you please rebase this pr? :) Seems all the service account changes are in here as well.

andreasgerstmayr:
Sure, done. I think GitHub usually doesn't show commits which already exist in the target branch, however in this case due to the squash commit the commits didn't match up anymore. But anyway, I think squashing of PRs is an elegant way to keep the history tidy, so let's not change that ;)

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -142,7 +142,7 @@ type ObjectStorageSpec struct {
 	// Name of a secret in the same namespace as the tempo Microservices custom resource.
 	//
 	// +optional
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
andreasgerstmayr, line 145: To be honest I'm not sure if this marker changed anything at all (?)
Afaics it should have made the validation of an empty secret name obsolete (but it didn't).

pavolloffay (approved):

frzifus (approved):

PR #129, opened 2022-12-06
Expose http port of querier and distributor

andreasgerstmayr:


andreasgerstmayr:
@pavolloffay to gather metrics from these components. Do you recommend to conditionally expose these ports (possibly with a config option in the CR) or not at all?

andreasgerstmayr:
hm, afaics a PodMonitor would also work, then we don't need to expose these ports in the service

pavolloffay (approved):

frzifus (approved):

PR #146, opened 2022-12-22
Mark storage.secret required in the CRD

andreasgerstmayr:


frzifus (approved):
oh actually i thought we did this already :D

andreasgerstmayr:
> oh actually i thought we did this already :D

Yep, we did check for existence of the secret in the validating webhook already.
This PR adds validation properties to the CRD (`required` and `minLength: 1`), so tools can perform basic validations before calling the webhook.

Issue #133, opened 2022-12-12
Watch the storage secret and regenerate the tempo config if required


andreasgerstmayr:
The operator should watch the storage secret and regenerate the tempo configuration if the secret changes.

andreasgerstmayr:
Ref. https://book.kubebuilder.io/reference/watching-resources/externally-managed.html

PR #154, opened 2023-01-05
Watch storage secret and reconcile on changes

andreasgerstmayr:
The tempo configuration file (ConfigMap) needs to be updated if the content of the storage secret changes.

Resolves: #133

### DIFF
--- a/controllers/tempo/microservices_controller.go
+++ b/controllers/tempo/microservices_controller.go
@@ -9,14 +9,24 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/os-observability/tempo-operator/apis/tempo/v1alpha1"
 	"github.com/os-observability/tempo-operator/internal/manifests"
+)
+
+const (
+	storageSecretField = ".spec.storage.secret" // nolint #nosec
 )
 
 // MicroservicesReconciler reconciles a Microservices object.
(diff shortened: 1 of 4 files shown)


tests/e2e/reconcile/04-check-bucket-name.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # workaround for asserting parts of the configmap
+  # can be removed once https://github.com/kudobuilder/kuttl/issues/262 is implemented
+  - command: "/bin/sh -c 'kubectl get --namespace $NAMESPACE configmap tempo-simplest -o yaml | grep \"bucket: tempo-updated\"'"
```
andreasgerstmayr, line 6: Alternatively we could assert the entire configmap here, but then this tests needs updating for every (unrelated) change to the configmap.

controllers/tempo/microservices_controller.go
```diff
@@ -132,13 +142,32 @@ func (r *MicroservicesReconciler) getStorageConfig(ctx context.Context, tempo v1
...
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.Microservices{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&corev1.Service{}).
 		Owns(&v1.StatefulSet{}).
 		Owns(&v1.Deployment{}).
+		Watches(
+			&source.Kind{Type: &corev1.Secret{}},
```
andreasgerstmayr, line 167: I'm not 100% happy that this code watches the entire cluster for every change to any secret, but afaics that's how it's supposed to work when the operator needs to react on changes of external resources: https://book.kubebuilder.io/reference/watching-resources/externally-managed.html

controllers/tempo/microservices_controller.go
```diff
@@ -150,3 +179,26 @@ func isNamespaceScoped(obj client.Object) bool {
 		return true
 	}
 }
+
+func (r *MicroservicesReconciler) findMicroservicesForStorageSecret(secret client.Object) []reconcile.Request {
+	microservices := &v1alpha1.MicroservicesList{}
+	listOps := &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(storageSecretField, secret.GetName()),
```
pavolloffay, line 186: this is quite nice to query based on a field value

controllers/tempo/microservices_controller.go
```diff
@@ -150,3 +179,26 @@ func isNamespaceScoped(obj client.Object) bool {
...
+		Namespace:     secret.GetNamespace(),
+	}
+	err := r.List(context.TODO(), microservices, listOps)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+
+	requests := make([]reconcile.Request, len(microservices.Items))
+	for i, item := range microservices.Items {
+		requests[i] = reconcile.Request{
```
pavolloffay, line 196: Should we consider checking for changes here and trigger reconciling only if a valid change is detected?

andreasgerstmayr: My envisioned workflow of this would be that on every secret change, the Reconcile function validates the secret and if the validation fails adds these errors in the status field (https://github.com/os-observability/tempo-operator/issues/153). If the validation succeeds, it updates the ConfigMap.

If relevant parts of the secret changed, the hash of the ConfigMap will be different and pods are restarted (https://github.com/os-observability/tempo-operator/issues/137). If no relevant parts of the secret changed (e.g. a user added a new key to the secret, which is not read by the operator), the checksum of the ConfigMap will stay the same and the pods are not restarted.

andreasgerstmayr: (i.e. if the validation fails, Reconcile should still run and the error `Condition` should be added to the status field)

controllers/tempo/microservices_controller.go
```diff
@@ -132,13 +142,32 @@ func (r *MicroservicesReconciler) getStorageConfig(ctx context.Context, tempo v1
...
 func (r *MicroservicesReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Add an index to the storage secret field in the Microservices CRD.
+	// If the content of any secret changes, the watcher can identify related Microservices CRs
+	// and reconcile them (i.e. update the tempo configuration file and restart the pods)
+	err := mgr.GetFieldIndexer().IndexField(context.Background(), &v1alpha1.Microservices{}, storageSecretField, func(rawObj client.Object) []string {
+		microservices := rawObj.(*v1alpha1.Microservices)
+		if microservices.Spec.Storage.Secret == "" {
+			return nil
+		}
+		return []string{microservices.Spec.Storage.Secret}
```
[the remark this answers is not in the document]
andreasgerstmayr: afaics no. I just tested the case of having two Microservices CRs in the same namespace referencing the same secret, and it works as expected (when the secret gets updated, a reconcile request is created for each CR).

I noticed another thing though (unrelated to above), that EnqueueRequestsFromMapFunc() calls the handler twice on updates: once with the old object (Secret in our case) and once with the new object:
https://github.com/kubernetes-sigs/controller-runtime/blob/9fd57745d6e93f85ee59e2a3b17f3d84dc5ef688/pkg/handler/enqueue_mapped.go#L61-L65 - not sure why.

pavolloffay (approved):

iblancasa (approved):

Issue #137, opened 2022-12-16
Own the tempo ConfigMap and restart pods on changes


andreasgerstmayr:
The MicroservicesController should `Own()` the tempo ConfigMap and restart pods if it changes.

andreasgerstmayr:
A way to restart pods on ConfigMap changes:
Store checksum of ConfigMap contents in an annotation: https://blog.questionable.services/article/kubernetes-deployments-configmap-change/
Example in Loki operator: https://github.com/grafana/loki/blob/d086c237dfd6d70d53bc16dffef6b0ab713a366b/operator/internal/manifests/distributor.go#L122

PR #158, opened 2023-01-12
Restart pods when the tempo configuration changes

andreasgerstmayr:
This change adds the checksum of the tempo configuration file to the PodTemplateSpec of each Deployment and StatefulSet. If the tempo configuration changes, the checksum will be updated, and the Deployments and StatefulSet will trigger a recreation of the pods.
This is required for tempo to load the new configuration.

Resolves: #137
Note: This PR is based on #154 because of the e2e tests, I'll rebase this PR once #154 is merged.

### DIFF
--- a/internal/manifests/config/configmap.go
+++ b/internal/manifests/config/configmap.go
@@ -25,6 +25,9 @@ type S3 struct {
 	Bucket   string
 }
 
+// BuildConfigMap builds the tempo configuration file and the tenant-specific overrides configuration.
+// It returns a ConfigMap containing both configuration files and the checksum of the main configuration file
+// (the tenant-specific configuration gets reloaded automatically, therefore no checksum is required).
 func BuildConfigMap(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMap, string, error) {
 	config, err := buildConfiguration(tempo, params)
 	if err != nil {
--- a/internal/manifests/config/configmap.go
+++ b/internal/manifests/config/configmap.go
@@ -1,6 +1,9 @@
 package config
 
 import (
+	"crypto/sha256"
+	"fmt"
+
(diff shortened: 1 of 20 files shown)


andreasgerstmayr:
rebased

internal/manifests/config/configmap.go
```diff
@@ -22,15 +25,15 @@ type S3 struct {
 	Bucket   string
 }
 
-func BuildConfigMap(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMap, error) {
+func BuildConfigMap(tempo v1alpha1.Microservices, params Params) (*corev1.ConfigMap, string, error) {
```
frzifus, line 28: Maybe we could put a comment on top that describes whats actually returned. wdyt?

internal/manifests/manifestutils/params.go
```diff
@@ -0,0 +1,22 @@
+package manifestutils
```
frzifus, line 1: Could it lead to some issues, that this params are in an `internal` path? E.g. if some other peace of software wants to import the CR?

andreasgerstmayr: afaics the params are only used in `manifests.BuildAll` and all functions building manifests, for example `BuildCompactor` etc (which is located in `internal/`). Previously `Params` was defined in `internal/manifests/manifests.go`, I've moved it to `internal/manifests/manifestutils/params.go` (due to having a cyclic import otherwise).

The CR doesn't reference this struct, it's mainly for passing around params to the manifest-build-functions.

internal/manifests/config/configmap.go
andreasgerstmayr: Good idea, done.

frzifus (approved):

PR #159, opened 2023-01-15
Protect internal components using TLS

rubenvp8510:
This PR adds the capability to protect the internal components using TLS.

This approach generated the CA, CABundle, and all the certs/private keys for each components using GoLang. It first created all the secrets, and the CABundle `ConfigMap`.

Then it configure each component , mounting each secret as a volumes in the different components, also mounting the CA bundle.

There is a cert rotation controller that is in charge of monitoring the certificates expiration, and annotate the secrets and the `Microservice` if a new certificate is required due an expiration.

This approach was taken from the Loki operator, almost all the `certrotation` package is copied from there, this package is in charge of create the CA, the bundle and all the certificates and private keys, create the `ConfigMap` and secret objects with he corresponding annotations

The buildIn cert manager could be enabled/disabled in  ProjectConfig using some FeatureFlags. I added some comments on how that should operate.

In this PR I didn't set the `minTLS` version and the allowed ciphers, I would leave this for another PR as this is large enough. 

Create a shared library to cert-manager https://github.com/grafana/loki/issues/8181


tests/e2e/reconcile/01-install.yaml
```diff
@@ -15,6 +15,8 @@ kind: Microservices
 metadata:
   name: simplest
 spec:
+  images:
+    tempoQuery: quay.io/rvargasp/tempo-query
```
iblancasa, line 19: What does this image contain?

internal/certrotation/var.go
```diff
@@ -0,0 +1,47 @@
...
+func SigningCASecretName(stackName string) string {
+	return fmt.Sprintf("tempo-%s-signing-ca", stackName)
+}
+
+// CABundleName returns the microservices ca bundle configmap name.
+func CABundleName(stackName string) string {
+	return fmt.Sprintf("tempo-%s-ca-bundle", stackName)
+}
+
+// ComponentCertSecretNames retruns a list of all tempo component certificate secret names.
```
iblancasa, line 33:
```suggestion
// ComponentCertSecretNames returns a list of all tempo component certificate secret names.
```

internal/certrotation/rotation_test.go
```diff
@@ -0,0 +1,335 @@
...
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, issuer, requestKey, issuerKey)
+	if err != nil {
+		return nil, err
+	}
+	certs, err := x509.ParseCertificates(derBytes)
+	if err != nil {
+		return nil, err
+	}
+	if len(certs) != 1 {
+		return nil, errors.New("Expected a single certificate")
```
iblancasa, line 332:
```suggestion
return nil, fmt.Errorf("Expected a single certificate. Found: %s", len(certs))
```

internal/certrotation/rotation_test.go
```diff
@@ -0,0 +1,335 @@
...
+				return nowCA, nil
+			},
+			refresh:    2 * time.Minute,
+			wantReason: "already expired",
+		},
+		{
+			desc: "refresh only when expired",
+			annotations: map[string]string{
+				CertificateIssuer:              "creator-tests",
+				CertificateNotAfterAnnotation:  now.Add(45 * time.Minute).Format(time.RFC3339),
```
iblancasa, line 198: Is there any reason for `45`?

iblancasa (changes requested):

tests/e2e/reconcile/01-install.yaml
rubenvp8510: that image contains the jaegerQuery pluign that supports TLS communication with the tempo front-end. We can remove this when the new version of the oficial image is released.

internal/certrotation/rotation_test.go
rubenvp8510: No reason, is just a test, needs to put a number.

bundle/manifests/tempo-operator-manager-config_v1_configmap.yaml
```diff
@@ -25,6 +25,19 @@ data:
...
+      grpcEncryption: true
+      builtInCertManagement:
+        enabled: true
+        # CA certificate validity: 5 years
+        caValidity: 43830h
+        # CA certificate refresh at 80% of validity
+        caRefresh: 35064h
+        # Target certificate validity: 90d
+        certValidity: 2160h
+        # Target certificate refresh at 80% of validity
```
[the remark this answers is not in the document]
rubenvp8510: This comes from Loki approach, I think this is to be safe and make sure we always have new certificates before those expired.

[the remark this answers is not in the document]
iblancasa: Ok, thanks for the explanation!

tests/e2e/reconcile/01-install.yaml
iblancasa: Could you add a comment explaining that? Even, I would create an issue in this repository to do that change in the future.

internal/certrotation/var.go
```diff
@@ -0,0 +1,47 @@
...
+
+// CABundleName returns the microservices ca bundle configmap name.
+func CABundleName(stackName string) string {
+	return fmt.Sprintf("tempo-%s-ca-bundle", stackName)
+}
+
+// ComponentCertSecretNames returns a list of all tempo component certificate secret names.
+func ComponentCertSecretNames(stackName string) []string {
+	return []string{
+		fmt.Sprintf("tempo-%s-distributor-http", stackName),
```
andreasgerstmayr, line 36: What's the reason for creating a different cert for each port of each service?
I was looking at the loki operator, and it creates a new service for each port (https://github.com/grafana/loki/blob/61c02d78f2f54c00c6aa47943c4944dc7b0619a7/operator/internal/manifests/distributor.go#L153-L208), but I don't know why.

internal/manifests/config/var.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+import "fmt"
+
+func fqdn(serviceName, namespace string) string {
+	return fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, namespace)
```
andreasgerstmayr, line 6: I'm not sure, does this need to be configurable? https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/ provides a `--service-dns-domain` option (defaults to `cluster.local`).

internal/manifests/manifests.go
```diff
@@ -37,26 +39,32 @@ func BuildAll(params Params) ([]client.Object, error) {
...
 		Bucket:   params.StorageParams.S3.Bucket,
-	}})
+	}, HTTPEncryption: params.Gates.HTTPEncryption,
+		GRPCEncryption: params.Gates.GRPCEncryption})
 	if err != nil {
 		return nil, err
 	}
 
-	ingesterObjs, err := ingester.BuildIngester(params.Tempo)
+	ingesterObjs, err := ingester.BuildIngester(params.Tempo, params.Gates)
```
andreasgerstmayr, line 48: In #158 I had to pass down the tempo configuration file checksum to each pod, so I followed Loki's approach to pass it down in the `param` struct. I think we could also add the `Gates` field to the `params` struct, and pass it down to all `Build*` functions instead of passing each var separately, wdyt?

andreasgerstmayr (commented):
Looks great overall, I've added a couple of minor comments/suggestions.

iblancasa (approved):

internal/manifests/manifests.go
rubenvp8510: yes, I think it makes more sense, I'll change this to pass all the params.

internal/certrotation/var.go
rubenvp8510: I think the idea is to mantain the HTTP and gRPC TLS configurations separate, as int could be enabled or disables by independent flags. but on the other side, you're right, may be we can use the same certs. I need to think a little bit more on this.

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -7,6 +7,59 @@ import (
 	apiv1alpha1 "github.com/os-observability/tempo-operator/apis/tempo/v1alpha1"
 )
 
+// BuiltInCertManagement is the configuration for the built-in facility to generate and rotate
+// TLS client and serving certificates for all Tempo services and internal clients except
+// for the tempo-gateway.
+type BuiltInCertManagement struct {
+	// CACertValidity defines the total duration of the CA certificate validity.
+	CACertValidity string `json:"caValidity,omitempty"`
```
pavolloffay, line 15: does the type represent duration? Could it be changed to use (golang/k8s) duration?

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -7,6 +7,59 @@ import (
 	apiv1alpha1 "github.com/os-observability/tempo-operator/apis/tempo/v1alpha1"
 )
 
+// BuiltInCertManagement is the configuration for the built-in facility to generate and rotate
+// TLS client and serving certificates for all Tempo services and internal clients except
+// for the tempo-gateway.
+type BuiltInCertManagement struct {
+	// CACertValidity defines the total duration of the CA certificate validity.
```
pavolloffay, line 14: what is the default value?

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -7,6 +7,59 @@ import (
...
+	CACertValidity string `json:"caValidity,omitempty"`
+	// CACertRefresh defines the duration of the CA certificate validity until a rotation
+	// should happen. It can be set up to 80% of CA certificate validity or equal to the
+	// CA certificate validity. Latter should be used only for rotating only when expired.
+	CACertRefresh string `json:"caRefresh,omitempty"`
```
pavolloffay, line 19: The same comment as previous one:
* could a duration be used as a type? 
* what is the default? 

also the same comment applies for other fields in this file.

internal/manifests/manifestutils/service.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"fmt"
+	"path"
+
+	"github.com/ViaQ/logerr/v2/kverrors"
+	"github.com/imdario/mergo"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TempoServerGRPCTLSDir() string {
```
pavolloffay, line 12: please add godoc to public APIs

tests/e2e/smoketest-with-jaeger/01-install.yaml
```diff
@@ -15,6 +15,8 @@ kind: Microservices
 metadata:
   name: simplest
 spec:
+  images:
+    tempoQuery: quay.io/rvargasp/tempo-query
```
pavolloffay, line 19: the same here add a comment and link to an issue to revert it back

pavolloffay (commented):
LGTM, just some minor comments.

How is handled the use case when a secret with the cert is removed by user? The jaeger-operator is caching the certs on its filesystem and reconstracts the secret.

apis/config/v1alpha1/projectconfig_types.go
rubenvp8510: Yes,I'll do the change.

apis/config/v1alpha1/projectconfig_types.go
rubenvp8510: The default value is what is provided on the `controller_manager_config` actually I think if you don't set anything it will be zero.

rubenvp8510: > LGTM, just some minor comments.
> 
> How is handled the use case when a secret with the cert is removed by user? The jaeger-operator is caching the certs on its filesystem and reconstracts the secret.



> LGTM, just some minor comments.
> 
> How is handled the use case when a secret with the cert is removed by user? The jaeger-operator is caching the certs on its filesystem and reconstracts the secret.



> LGTM, just some minor comments.
> 
> How is handled the use case when a secret with the cert is removed by user? The jaeger-operator is caching the certs on its filesystem and reconstracts the secret.

The certificates will be regenerated using the same CA and the parameters of the expiration, host name etc..

rubenvp8510:
Because we already added readiness probe, I had to send a PR to tempo with this https://github.com/grafana/tempo/pull/2045, so this PR is dependent from that other one.

Because the readiness probe will be protected  by mTLS and kubetnetes doesn't support that, I exposed the ready endpoint in another port, same approach Loki did,  I'm using a custom image here. If this PR get merged we need to take that into account.

rubenvp8510:
The PR is ready for review, but there is an small thing, I haven't found a way to protect the communication between the `query <---> query-frontend` without enable TLS on the `query-frontend` endpoint, the bad thing for this is that if Jaeger is not enabled the query-frontend will be the public facing service and it will be protected by the internal TLS certs. I think we can disable the `HTTP Encryption` until we solve it, In Loki they have the gateway and is the gateway what is exposed.

frzifus (commented):
Since its blocked for a while, can we disable this feature and merge? :b

internal/manifests/manifests.go
```diff
@@ -15,12 +17,31 @@ import (
 	"github.com/os-observability/tempo-operator/internal/manifests/serviceaccount"
 )
 
+// Params holds parameters used to create Tempo objects.
+type Params struct {
```
andreasgerstmayr, line 21: Can you use the `Params` struct from https://github.com/os-observability/tempo-operator/blob/4c03363347a3ae70f3272accd117fa7febaa8734/internal/manifests/manifestutils/params.go? It's identical, only moved to another package due to a circular dependency.

andreasgerstmayr (approved):
I've added one comment regarding the `Params` struct.
Looks good overall, some merge conflicts need to be resolved before being able to merge this PR.

frzifus (approved):
Do we need here another follow-up ticket to include the tempo gateway?

controllers/tempo/microservices_controller.go
```diff
@@ -118,7 +128,7 @@ func (r *MicroservicesReconciler) reconcileManifests(ctx context.Context, log lo
 		}
 	}
 
-	objects, err := manifests.BuildAll(manifestutils.Params{Tempo: tempo, StorageParams: *storageConfig})
+	objects, err := manifests.BuildAll(manifestutils.Params{Tempo: tempo, StorageParams: *storageConfig, Gates: r.FeatureGates})
```
pavolloffay, line 131: shall we have a unit test to check if certs are created if TLS is enabled?

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -7,6 +7,59 @@ import (
...
+	// - `service-ca.crt`: The CA signing the service certificate in `tls.crt`.
+	BuiltInCertManagement BuiltInCertManagement `json:"builtInCertManagement,omitempty"`
+	// HTTPEncryption enables TLS encryption for all HTTP Microservices services.
+	// Each HTTP service requires a secret named as the service with the following data:
+	// - `tls.crt`: The TLS server side certificate.
+	// - `tls.key`: The TLS key for server-side encryption.
+	// In addition each service requires a configmap named as the Microservices CR with the
+	// suffix `-ca-bundle`, e.g. `tempo-dev-ca-bundle` and the following data:
+	// - `service-ca.crt`: The CA signing the service certificate in `tls.crt`.
+	HTTPEncryption bool `json:"httpEncryption,omitempty"`
```
pavolloffay, line 52: I think the comment should make it clear which services are protected and which not (public vs internal)

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -7,6 +7,59 @@ import (
...
+	// GRPCEncryption enables TLS encryption for all GRPC Microservices services.
+	// Each GRPC service requires a secret named as the service with the following data:
+	// - `tls.crt`: The TLS server side certificate.
+	// - `tls.key`: The TLS key for server-side encryption.
+	// In addition each service requires a configmap named as the Microservices CR with the
+	// suffix `-ca-bundle`, e.g. `tempo-dev-ca-bundle` and the following data:
+	// - `service-ca.crt`: The CA signing the service certificate in `tls.crt`.
+	GRPCEncryption bool `json:"grpcEncryption,omitempty"`
```
pavolloffay, line 60: the same as above

pavolloffay (commented):
overall LGTM just some minor comments. 


Could you please rebase it?

internal/certrotation/handlers/rotate_certs.go
```diff
@@ -0,0 +1,100 @@
...
+			continue
+		}
+
+		msg := fmt.Sprintf("Resource has been %s", op)
+		switch op {
+		case ctrlutil.OperationResultNone:
+			l.V(1).Info(msg)
+		case ctrlutil.OperationResultCreated:
+		case ctrlutil.OperationResultUpdated:
+		case ctrlutil.OperationResultUpdatedStatus:
```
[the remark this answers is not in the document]
rubenvp8510: I think we can do more log improvements in follow up PRS :)

internal/certrotation/handlers/options.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/ViaQ/logerr/v2/kverrors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/os-observability/tempo-operator/internal/certrotation"
+)
+
+//var serviceCAnnotationsRe = regexp.MustCompile(`^service.(?:alpha|beta)\.openshift\.io\/.+`)
```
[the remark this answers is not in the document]
rubenvp8510: Done!

controllers/tempo/microservices_controller.go
```diff
@@ -84,6 +87,13 @@ func (r *MicroservicesReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
+	if r.FeatureGates.BuiltInCertManagement.Enabled {
+		err := handlers.CreateOrRotateCertificates(ctx, log, req, r.Client, r.Scheme, r.FeatureGates)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("build in cert manager error: %w", err)
```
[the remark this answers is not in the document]
rubenvp8510: Changed, thanks nice catch!

controllers/tempo/microservices_controller.go
rubenvp8510: Yes, but due the size of this PR I would prefer to write more tests on follow up PR, if there is no objections I'll disable TLS until we have the tests and the right images in place. Also I would like to integrate tempo gateway to this.

pavolloffay (approved):

frzifus (approved):

frzifus:
ahh the e2e test failed. i triggered a rerun. ^^

Issue #153, opened 2023-01-05
Move storage secret validation to reconcile loop (status field)


andreasgerstmayr:
The validation of the storage secret should happen in the reconcile loop instead of the validation webhook of the CR, as the (external) secret can change at any time, but the validation webhook is only called when the CR changes.

If there is a validation error, a `Condition` should be set to the `status` field of the CR.

Ref:
https://github.com/grafana/loki/blob/fb3ad058e0de6bb4e3041691085df8a2c10efafb/operator/internal/handlers/lokistack_create_or_update.go#L84
https://github.com/grafana/loki/blob/fb3ad058e0de6bb4e3041691085df8a2c10efafb/operator/controllers/loki/lokistack_controller.go#L145
https://github.com/grafana/loki/blob/fb3ad058e0de6bb4e3041691085df8a2c10efafb/operator/internal/status/lokistack.go#L183-L189

PR #167, opened 2023-01-19
Validate storage secret in reconcile loop

andreasgerstmayr:
Set Ready or Degraded condition in the status field based on the current status of the external storage secret.

In the future, these conditions should be updated based on the overall health of the tempo instance.

Resolves: #153


bundle/manifests/tempo.grafana.com_microservices.yaml
```diff
@@ -580,6 +580,75 @@ spec:
           status:
             description: MicroservicesStatus defines the observed state of Microservices.
             properties:
+              conditions:
+                description: Conditions of the Tempo deployment health.
+                items:
+                  description: "Condition contains details for one aspect of the current
+                    state of this API Resource. --- This struct is intended for direct
```
andreasgerstmayr, line 587: Is there a way to ignore this description? This is coming from the comment in `[]metav1.Condition`.

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -80,11 +80,38 @@ type MicroservicesSpec struct {
...
+	ConditionDegraded ConditionStatus = "Degraded"
+)
+
+// ConditionReason defines possible reasons for each condition.
+type ConditionReason string
+
+const (
+	// ReasonReady defines a healthy tempo instance.
+	ReasonReady ConditionReason = "Ready"
+	// ReasonInvalidStorageConfig defines that the object storage configuration is invalid (missing or incomplete storage secret).
```
iblancasa, line 99: Just asking: would it make sense to differentiate between the two conditions? To provide more information in the log or something

bundle/manifests/tempo.grafana.com_microservices.yaml
iblancasa: It seems you're not the first one finding this "issue": https://docs.openshift.com/container-platform/4.11/rest_api/operatorhub_apis/operatorcondition-operators-coreos-com-v2.html#spec

controllers/tempo/microservices_controller.go
```diff
@@ -36,6 +41,16 @@ type MicroservicesReconciler struct {
...
 }
 
+type DegradedError struct {
+	Reason  v1alpha1.ConditionReason
+	Message string
+	Requeue bool
+}
+
+func (e *DegradedError) Error() string {
+	return fmt.Sprintf("cluster degraded: %s", e.Message)
```
iblancasa, line 51: How about adding the `Reason` to the message?

iblancasa (approved):

apis/tempo/v1alpha1/microservices_types.go
andreasgerstmayr: I don't follow, can you elaborate on that? Currently the condition is `Ready` if everything is working, and `Degraded` if something is broken. If it's `Degraded`, the `ConditionReason` will explain why it's `Degraded`.

controllers/tempo/microservices_controller.go
andreasgerstmayr: Updated.

controllers/tempo/microservices_controller_test.go
```diff
@@ -93,9 +102,124 @@ func TestReconcile(t *testing.T) {
...
+
+	// Verify status conditions: Ready=true
+	updatedTempo1 := v1alpha1.Microservices{}
+	err = k8sClient.Get(context.Background(), nsn, &updatedTempo1)
+	require.NoError(t, err)
+	assert.Equal(t, []metav1.Condition{{
+		Type:               "Ready",
+		Status:             "True",
+		LastTransitionTime: updatedTempo1.Status.Conditions[0].LastTransitionTime,
+		Reason:             "Ready",
```
rubenvp8510, line 150: May be we can use the conditionReasons constants defined in the `microservices_types.go` ?

rubenvp8510:
Small comment on the tests. but LGTM, we only need to solve the conflicts :D

controllers/tempo/microservices_controller_test.go
andreasgerstmayr: Good idea, I've updated the PR.

andreasgerstmayr:
rebased

apis/tempo/v1alpha1/microservices_types.go
iblancasa: I see. I was asking because, from the description of the status, it seems like there is actually more than one condition for the "Degraded" state:
> missing or incomplete storage secret

So, I was wondering if it would be a good idea to have something like `Missing` and `Incomplete Secret` instead of just `Degraded`.

iblancasa (approved):

apis/tempo/v1alpha1/microservices_types.go
andreasgerstmayr: Ah, now I understand (I thought "differentiate between the two conditions" was referring to `Ready` vs. `Degraded`, which got me confused).

> So, I was wondering if it would be a good idea to have something like `Missing` and `Incomplete Secret` instead of just `Degraded`.

Good question - incomplete secret could also have multiple reasons (`MissingEndpoint`, `InvalidEndpoint`, `MissingAccessKeyId`, `MissingAccessKeySecret`, etc).

In my opinion the `Reason` field is most useful if another operator reads this field and can react based on it (in this sense, I'd count it as part of the API which the tempo-operator exposes). In this case however I guess it'll be only administrators reacting to this error condition. Admins then (hopefully) also read the full error message (which gives further details).

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -140,58 +139,6 @@ func (v *validator) validateServiceAccount(ctx context.Context, tempo *Microserv
 	return allErrs
 }
 
-func (v *validator) validateStorageSecret(tempo *Microservices, storageSecret *corev1.Secret) field.ErrorList {
```
pavolloffay, line 143: Should we consider doing validation here as well and to catch "init errors"? 

I find it useful to inform the user that the config is not correct when creating the instance

controllers/tempo/microservices_controller.go
```diff
@@ -143,6 +174,12 @@ func (r *MicroservicesReconciler) getStorageConfig(ctx context.Context, tempo v1
 		return nil, fmt.Errorf("storage secret should contain endpoint and bucket, access_key_id and access_key_secret fields")
 	}
 
+	u, err := url.ParseRequestURI(string(storageSecret.Data["endpoint"]))
```
pavolloffay, line 177: it would be probably make sense to extract the validation of the secret into a separate function

pavolloffay (changes requested):

apis/tempo/v1alpha1/microservices_webhook.go
andreasgerstmayr: Can we guarantee that users always create the secret before the tempo CR?

An additional feature of this PR is that now users can apply the manifests in any order and it'll still work, while before the order was important (first apply the secret, then the CR). That's also how Pods works, if they reference a secret which doesn't exist the pod can still get created, but the Pod's status is `CreateContainerConfigError` and the `ContainersReady` condition is set to false.

pavolloffay: > Can we guarantee that users always create the secret before the tempo CR?

Good comment. I definitely like the implementation in this PR. However, I would still advise users to create the secret first. The error propagation in the validating webhook is just better in my opinion.

pavolloffay (approved):

Issue #166, opened 2023-01-19
Add generate command


andreasgerstmayr:
Add a new `generate` command to generate manifests from the CLI.

Input: Tempo CR
Output: YAML manifests of all Deployments, Services, etc.

PR #168, opened 2023-01-19
Implement generate command

andreasgerstmayr:
This debug utility reads a CR and outputs the manifests which a running instance of this operator would apply to the cluster.

Resolves: #166


cmd/generate/main.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return nil
+}
+
+func usage() {
+	println("usage: go run ./cmd/generate/main.go -cr path/to/microservices/cr.yaml")
+}
+
+func main() {
```
andreasgerstmayr, line 92: Alternatively we could also split up the main tempo-operator binary in multiple sub commands using a library like spf13/cobra.

pavolloffay: I would prefer that, it will be easier to use it (not additional build is required)

iblancasa (approved):
For the Jaeger Operator, there was an E2E test for this feature. Would it make sense to add something similar here?

cmd/generate/main.go
rubenvp8510: I agree with this, having a sub commands will be better.

rubenvp8510: @andreasgerstmayr are you going to do this change? WDYT?

andreasgerstmayr: @rubenvp8510 yes, it's almost done. I just need to update a few places where the tempo-operator executable is started (in the manifests and Makefile).
I'll also add an e2e test and send a PR tomorrow.

cmd/generate/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+func loadSpec(path string) (v1alpha1.Microservices, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return v1alpha1.Microservices{}, err
+	}
+	defer file.Close()
+
+	spec := v1alpha1.Microservices{}
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 8192)
```
frzifus, line 40: `8192` is a memory limiation? ^^

cmd/generate/main.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			return fmt.Errorf("error: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func usage() {
+	println("usage: go run ./cmd/generate/main.go -cr path/to/microservices/cr.yaml")
```
frzifus, line 89: Is there a reason to use the builtin `println` here?

```
 Current implementations provide several built-in functions useful during bootstrapping. These functions are documented for completeness but are not guaranteed to stay in the language. They do not return a result. 
```
https://go.dev/ref/spec#Bootstrapping

cmd/generate/main.go
andreasgerstmayr: Done:
```
$ go run main.go                      
Usage:
  tempo-operator [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  generate    Generate YAML manifests from a Tempo CR
  help        Help about any command
  start       Start the Tempo operator

Flags:
      --config string   The controller will load its initial configuration from this file. Omit this flag to use the default configuration values. Command-line flags override configuration from this file.
  -h, --help            help for tempo-operator

Use "tempo-operator [command] --help" for more information about a command.
```

:)

cmd/generate/main.go
andreasgerstmayr: Docs say `bufferSize determines how far into the stream the decoder will look to figure out whether this is a JSON stream (has whitespace followed by an open brace).`
Imho it shouldn't have to look very far to determine if the input file is a JSON or YAML.

cmd/generate/main.go
andreasgerstmayr: Ah I didn't realize it's not an official command. I've updated the PR to use sub commands, and removed this function.

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -33,13 +33,19 @@ var microserviceslog = logf.Log.WithName("microservices-resource")
...
 		For(r).
-		WithDefaulter(&defaulter{defaultImages: defaultImages}).
+		WithDefaulter(NewDefaulter(defaultImages)).
 		WithValidator(&validator{client: mgr.GetClient()}).
 		Complete()
 }
 
 //+kubebuilder:webhook:path=/mutate-tempo-grafana-com-v1alpha1-microservices,mutating=true,failurePolicy=fail,sideEffects=None,groups=tempo.grafana.com,resources=microservices,verbs=create;update,versions=v1alpha1,name=mmicroservices.kb.io,admissionReviewVersions=v1
 
+func NewDefaulter(defaultImages ImagesSpec) *defaulter {
```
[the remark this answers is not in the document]
andreasgerstmayr: The godoc check is disabled by default in golang-lint: https://github.com/golangci/golangci-lint/issues/456

I've added a comment and exported the struct now.

andreasgerstmayr:
I've moved the command to a subcommand now, and added an e2e test.
Unfortunately the e2e test currently fails on Kubernetes 1.21 because it cannot parse the `availableReplicas` field in `status` of `v1.StatefulSet`.

Is there a better way to drop the status field other than marshalling the object, then unmarshalling to a map, removing the field, and marshalling again (to YAML)? :confused:

iblancasa:
> Unfortunately the e2e test currently fails on Kubernetes 1.21 because it cannot parse the availableReplicas field in status of v1.StatefulSet.

So, instead of matching the `StatefulSet`, how about matching the `Pods` managed by that `StatefulSet`?

iblancasa (changes requested):

tests/e2e/generate/cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: tempo.grafana.com/v1alpha1
```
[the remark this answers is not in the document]
andreasgerstmayr: This is the input CR for the generate command in this e2e test. Basically like https://github.com/jaegertracing/jaeger-operator/blob/main/tests/e2e/generate/generate/jaeger-template.yaml.template, but for tempo.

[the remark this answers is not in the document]
iblancasa: Oh, yes. Sorry. I didn't see the call in the previous file correctly.

cmd/generate/main.go
rubenvp8510: Can we put this number in a constant? and a comment indicating what it is?

cmd/generate/main_test.go
```diff
@@ -0,0 +1,65 @@
...
+		},
+		Data: map[string]string{
+			"tempo.yaml": "ingester:\n  setting: a\n",
+		},
+	}
+
+	var buf bytes.Buffer
+	err := toYAML(scheme, []client.Object{&cm}, &buf)
+	require.NoError(t, err)
+	require.Equal(t, `---
```
rubenvp8510, line 54: woth to use `YAMLEq` instead?

rubenvp8510 (approved):
Minor comments LGTM

cmd/generate/main_test.go
andreasgerstmayr: Good idea, I've updated the PR.

cmd/generate/main.go
andreasgerstmayr: Done.

andreasgerstmayr:
> > Unfortunately the e2e test currently fails on Kubernetes 1.21 because it cannot parse the availableReplicas field in status of v1.StatefulSet.
> 
> So, instead of matching the `StatefulSet`, how about matching the `Pods` managed by that `StatefulSet`?

The problem was that the generated YAML contained the status field, not the assertion in kuttl.

I took a look what `json.NewYAMLSerializer#Encode` does behind the scenes, and it actually also runs `json.Marshal` -> `yaml.Unmarshal` -> `yaml.Marshal`. `json.Marshal` is required to respect the json tags in the structs, and `yaml.Unmarshal` and `yaml.Marshal` to convert JSON to YAML. I've implemented the same now in `generate/main.go`, and remove the `status` field before the last step.

iblancasa:
> > > Unfortunately the e2e test currently fails on Kubernetes 1.21 because it cannot parse the availableReplicas field in status of v1.StatefulSet.
> > 
> > 
> > So, instead of matching the `StatefulSet`, how about matching the `Pods` managed by that `StatefulSet`?
> 
> The problem was that the generated YAML contained the status field, not the assertion in kuttl.
> 
> I took a look what `json.NewYAMLSerializer#Encode` does behind the scenes, and it actually also runs `json.Marshal` -> `yaml.Unmarshal` -> `yaml.Marshal`. `json.Marshal` is required to respect the json tags in the structs, and `yaml.Unmarshal` and `yaml.Marshal` to convert JSON to YAML. I've implemented the same now in `generate/main.go`, and remove the `status` field before the last step.

Awesome! 🚀

iblancasa (approved):

frzifus (approved):

Issue #149, opened 2022-12-22
Configure readiness probes


andreasgerstmayr:
See https://github.com/grafana/tempo/blob/0ed9ddcdbb38f2a6348f665ccd52aea3dc7a3b8e/operations/jsonnet-compiled/Deployment-querier.yaml#L39-L44 for example

PR #174, opened 2023-01-25
Configure readiness probes

andreasgerstmayr:
Resolves: #149

### DIFF
--- a/internal/manifests/manifestutils/probes.go
+++ b/internal/manifests/manifestutils/probes.go
@@ -1,0 +1,20 @@
+package manifestutils
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// TempoReadinessProbe returns a readiness Probe spec for tempo components.
+func TempoReadinessProbe() *corev1.Probe {
+	return &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: TempoReadinessPath,
+				Port: intstr.FromString(HttpPortName),
+			},
+		},
+		InitialDelaySeconds: 15,
+		TimeoutSeconds:      1,
+	}
+}
--- a/tests/e2e/reconcile/01-install.yaml
+++ b/tests/e2e/reconcile/01-install.yaml
@@ -18,16 +18,3 @@ spec:
   storage:
     secret: minio-test
   storageSize: 200M
-  resources:
-    total:
-      limits:
-        memory: 2Gi
-        cpu: 2000m
-  template:
-    queryFrontend:
-      jaegerQuery:
-        enabled: true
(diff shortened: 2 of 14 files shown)


andreasgerstmayr:
e2e tests work fine locally. Afaics the problem is that the GitHub Actions VM has only 2 CPU cores, and because kuttl runs tests in parallel it can lead to pods stuck in pending state because of unavailable CPU resources.

Problem could be fixed by running e2e tests serially, but maybe there's a better way.

andreasgerstmayr:
> but maybe there's a better way.

and there is! I misread the message from the deployment and thought it's missing actual/physical (emulated) CPU cores, but in reality the *defined* `requests.cpu` of the pods were exceeding the calculated maximum (based on the CPU cores of the VM).

Easy fix is to deploy the CR without `resources` set.

iblancasa (approved):

internal/manifests/manifestutils/probes.go
```diff
@@ -0,0 +1,20 @@
...
+// TempoReadinessProbe returns a readiness Probe spec for tempo components.
+func TempoReadinessProbe() *corev1.Probe {
+	return &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: TempoReadinessPath,
+				Port: intstr.FromString(HttpPortName),
+			},
+		},
+		InitialDelaySeconds: 15,
```
rubenvp8510, line 17: are those fine hard coded? or should be allow some configuration here?

tests/e2e/reconcile/01-install.yaml
```diff
@@ -18,16 +18,3 @@ spec:
   storage:
     secret: minio-test
   storageSize: 200M
-  resources:
```
rubenvp8510, line 21: why removed this?

andreasgerstmayr: Because kuttl runs e2e tests in parallel, and if there are too many tempo containers running, each requesting a specific amount of CPU resources, the kubernetes scheduler will stop scheduling them (because the GitHub Actions VM only has 2 cores), pods will be in pending state and the e2e test will fail.

internal/manifests/manifestutils/probes.go
andreasgerstmayr: I think it's fine hard coded, these values mean that kubernetes waits for 15 seconds until it starts checking the readiness of the pod, and uses a timeout of 1 second for the readiness probe.

In case these values make issues later we can make them configurable, but I'm a bit reluctant to add config options for every tiny knob. wdyt?

rubenvp8510: It is fine to me. I think we can proceed in this way

rubenvp8510 (approved):

andreasgerstmayr:
rebased due to merge conflicts

PR #176, opened 2023-01-26
Add replication factor support

rubenvp8510:

This PR adds `replicationFactor`, and replicas for _ingester_ and distributor,

The  number of available _ingesters_ should be at least `floor(replicationFactor/2) + 1 `this means that replicas less than that could be considered invalid configuration. We can have `replicas = floor(replicationFactor/2) + 1` but with that configuration we can't lose _ingester_ replicas, anyway even with that inconvenient, is a valid configuration and tempo will work as long as no _ingester_ replica fails.

distributor replicas are not affected by the `replicationFactor`. the default values for both were set to 1.


apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -192,6 +211,23 @@ func (v *validator) validateStorage(ctx context.Context, tempo *Microservices) f
 	return v.validateStorageSecret(tempo, storageSecret)
 }
 
+func (v *validator) validateReplicationFactor(tempo *Microservices) field.ErrorList {
```
iblancasa, line 214: How about logging something in the operator when the replication factor is not valid?

iblancasa (changes requested):
I think this should be covered with an E2E test

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -98,6 +100,24 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
...
+
+	// Default replicas for ingester
+	if r.Spec.Components.Ingester == nil {
+		r.Spec.Components.Ingester = &TempoComponentSpec{}
+	}
+
+	if r.Spec.Components.Ingester.Replicas == nil {
+		r.Spec.Components.Ingester.Replicas = pointer.Int32(1)
+	}
+
```
andreasgerstmayr, line 120: Should we also set `r.Spec.ReplicationFactor` to the default value of `1` here?

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -192,6 +211,23 @@ func (v *validator) validateStorage(ctx context.Context, tempo *Microservices) f
...
 }
 
+func (v *validator) validateReplicationFactor(tempo *Microservices) field.ErrorList {
+	// Validate minimum quorum on ingestors according to replicas and replication factor
+	replicatonFactor := tempo.Spec.ReplicationFactor
+	ingesterReplicas := tempo.Spec.Components.Ingester.Replicas
+	quorum := int32(math.Floor(float64(replicatonFactor)/2.0) + 1)
+	// if ingester replicas less than quorum (which depends on replication factor), then doesn't allow to deploy as it is an
+	// invalid configuration. Quorum equal to replicas doesn't allow you to lose ingesters but is a valid configuration.
+	if (ingesterReplicas == nil && quorum > 1) || *ingesterReplicas < quorum {
```
andreasgerstmayr, line 221: If the default webhook is run, `tempo.Spec.Components.Ingester.Replicas` can never be `nil` here, right?
I'm wondering, is it possible that the default webhook doesn't run but the validation webhook runs?

apis/tempo/v1alpha1/microservices_webhook.go
andreasgerstmayr: Afaics the user who edits the CR gets immediate feedback if the validation fails (`field.ErrorList`) and is not allowed to save the CR, so it's not possible for the user to overlook any validation error.

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -98,6 +100,24 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 		r.Spec.SearchSpec.DefaultResultLimit = &defaultDefaultResultLimit
 	}
 
+	// Default replicas for distibutor
+	if r.Spec.Components.Distributor == nil {
+		r.Spec.Components.Distributor = &TempoComponentSpec{}
+	}
+
+	if r.Spec.Components.Distributor.Replicas == nil {
+		r.Spec.Components.Distributor.Replicas = pointer.Int32(1)
```
pavolloffay, line 109: maybe we could define a var and reuse it for all repliacs

apis/tempo/v1alpha1/microservices_webhook.go
```diff
@@ -98,6 +100,24 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 		r.Spec.SearchSpec.DefaultResultLimit = &defaultDefaultResultLimit
 	}
 
+	// Default replicas for distibutor
+	if r.Spec.Components.Distributor == nil {
```
pavolloffay, line 104: Could we change the `.Spec.Components.Distributor` (and friends) to a non-pointer struct?

rubenvp8510: I would be good!! great suggestion

apis/tempo/v1alpha1/microservices_webhook.go
rubenvp8510: No, I did this only because double check and precautions, but I'll change those things to not be pointers.

apis/tempo/v1alpha1/microservices_webhook.go
rubenvp8510: oh yeah! good catch

apis/tempo/v1alpha1/microservices_webhook.go
rubenvp8510: > I think this should be covered with an E2E test

Wondering if we need to to an E2E test of this, as I'm not sure how to capture webhook errors on kuttl, is not enough to have a unit test for our webhook validator?

rubenvp8510:
> I think this should be covered with an E2E test

Wondering if we need to do an E2E test for this, I'm not sure how to capture the webhook errors in a reliable way, Is not enought with the webhook validator unit tests?

andreasgerstmayr (approved):

iblancasa:
> > I think this should be covered with an E2E test
> 
> Wondering if we need to do an E2E test for this, I'm not sure how to capture the webhook errors in a reliable way using kuttl, Is not enough with the webhook validator unit tests?

Well, maybe. I don't know if that is enough.
From KUTTL you can capture it using a `TestStep` and check the `STDERR` from applying a manifest that will not pass the validation.

But I would not consider the absence of an E2E test a hard stopper to merge this PR.

iblancasa (approved):

Issue #140, opened 2022-12-16
query-frontend pod reconciliation doesn't work due to PodAntiAffinity role


andreasgerstmayr:
The PodAntiAffinity rule of the query-frontend pod prevents the deployment from adding a new pod (and terminating the old one) during reconciling if the cluster only has a single node: `Warning FailedScheduling 22s (x11 over 20m) default-scheduler 0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.`

Steps to reproduce:
* Create a CR
* Change a value triggering new deployments, for example modify `.spec.resources.total.limits.cpu`

PR #177, opened 2023-01-26
Update PodAntiAffinity for most components

andreasgerstmayr:
Using slightly modified best practices from the tempo-distributed helm chart at https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml.

In the helm chart
* distributor, query and query-frontend use podAntiAffinity with requiredDuringSchedulingIgnoredDuringExecution so that a pod cannot be scheduled on the same node twice, and preferredDuringSchedulingIgnoredDuringExecution for the failure domain
* ingester uses podAntiAffinity with preferredDuringSchedulingIgnoredDuringExecution so that a pod should not be scheduled on the same node or failure domain twice
* compactor has no podAntiAffinity set

In this PR distributor, query, query-frontend and ingester components get a podAntiAffinity with preferredDuringSchedulingIgnoredDuringExecution so that a pod should not be scheduled on the same node or failure domain twice. It's a "soft" affinity, i.e. if it cannot be satisfied (for example if the cluster only has a single node), the pod will still get scheduled (see #140).

Resolves: #140

### DIFF
--- a/internal/manifests/manifestutils/affinity.go
+++ b/internal/manifests/manifestutils/affinity.go
@@ -6,7 +6,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-// DefaultAffinity returns the default affinity for (most) of the Tempo components.
+// DefaultAffinity returns the default affinity for Tempo components.
 // It defines that two pods with the same labels (i.e. same component)
 // should not be scheduled on the same node or failure domain.
 func DefaultAffinity(labels labels.Set) *corev1.Affinity {
--- a/internal/manifests/manifestutils/affinity.go
+++ b/internal/manifests/manifestutils/affinity.go
@@ -1,0 +1,37 @@
+package manifestutils
+
+import (
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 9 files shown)


internal/manifests/manifestutils/affinity.go
```diff
@@ -0,0 +1,37 @@
+package manifestutils
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// DefaultAffinity returns the default affinity for (most) of the Tempo components.
```
pavolloffay, line 9: >  for (most) of the Tempo components.

I would remove this as this is a general API

pavolloffay (commented):
LGTM, just a minor comment

iblancasa (approved):

internal/manifests/manifestutils/affinity.go
andreasgerstmayr: Done

frzifus (approved):

Issue #183, opened 2023-01-30
Prepare the Tempo Operator to be released


iblancasa:
This task is to track the different changes needed to create a release of the Tempo Operator.
* [x] Add versioning to the code: add a `version` subcommand listing the versions of the different components. Also, provide a data structure to use the version in different parts of the code
* [*] ~Adapt the Dockerfile and Makefile to read the desired version to use for Tempo and Tempo Operator and build the source code correctly~
* [ ] ~Add automation to make easy the modification of the Tempo images and the generation of the bundle with the correct version~
* [ ] ~ocument all the manual steps that need to be performed~

pavolloffay:
This is dup/related to #3

pavolloffay:
@rubenvp8510 I am assiging it to you since you have already started. 

Based on #344 there are some missing pieces

- [ ] introduce GHA for changelog check and label for skipping the changelog
- [ ] Document how to find new versions of component images - maybe directly in `config/manager/controller_manager_config.yaml`

PR #184, opened 2023-01-30
Add versioning to the operator

iblancasa:

This PR is part of #183.

It adds the code to query the version in the Tempo Operator.


internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -239,7 +239,6 @@ func services(tempo v1alpha1.Microservices) []*corev1.Service {
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
-			ClusterIP: "None",
```
andreasgerstmayr, line 242: This line should not be deleted, this setting is required for the tempo cluster.

internal/manifests/queryfrontend/query_frontend_test.go
```diff
@@ -76,7 +76,6 @@ func getExpectedFrontendDiscoveryService(withJaeger bool) *corev1.Service {
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
-			ClusterIP: "None",
```
andreasgerstmayr, line 79: This line should not be deleted, this setting is required for the tempo cluster.

internal/version/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"runtime"
+)
+
+var (
+	version   string
+	buildDate string
+	tempo     string
```
andreasgerstmayr, line 11: The operator has a config to set the tempo image:
https://github.com/os-observability/tempo-operator/blob/3f19e127d4f608de628c14cc0f48c433540f8208/config/manager/controller_manager_config.yaml#L22-L24

and in the status field we parse the tag of the image: https://github.com/os-observability/tempo-operator/blob/3f19e127d4f608de628c14cc0f48c433540f8208/controllers/tempo/microservices_controller.go#L216-L221

I think we should also set the tempo version here based on the image.

internal/version/main.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// Version holds this Operator's version as well as the version of some of the components it uses.
+type Version struct {
+	Operator  string `json:"tempo-operator"`
```
andreasgerstmayr, line 16: As Tempo and Go are suffixed with `-version`, I'd also suffix the operator field with `-version`.

internal/version/main.go
```diff
@@ -0,0 +1,51 @@
...
+	BuildDate string `json:"build-date"`
```
andreasgerstmayr, line 17: Can you update the Makefile and Dockerfile to add the build date and operator version?
The git commit hash would also be helpful imho.

andreasgerstmayr (changes requested):

rubenvp8510:
I know may be not part of this PR but should we use the version as default image tag?

internal/manifests/queryfrontend/query_frontend.go
iblancasa: Good catch! I don't know why I removed it. Sorry

internal/version/main.go
iblancasa: This will be added in another PR (it is described in #183). I'm creating small PRs to make them easy to review

internal/version/main.go
iblancasa: Good suggestion! I'll implement it

andreasgerstmayr: Or should we make the version configurable in the CRD or the operator configuration, and set the docker image based on this (new) version field?

I'm not sure how configurable the version should be, as the operator will always depend on a specific version. The images field is handy if people use their own containers (which need to be compatible with the version X we tested the operator with).

main.go
```diff
@@ -6,12 +6,14 @@ import (
...
 	"github.com/os-observability/tempo-operator/cmd/generate"
 	"github.com/os-observability/tempo-operator/cmd/start"
+	"github.com/os-observability/tempo-operator/cmd/version"
 )
 
 func main() {
 	rootCmd := cmd.NewRootCommand()
 	rootCmd.AddCommand(start.NewStartCommand())
 	rootCmd.AddCommand(generate.NewGenerateCommand())
+	rootCmd.AddCommand(version.NewVersionCommand())
```
pavolloffay, line 16: How is the tempo version propagated to the version command? 

The tempo version can change and might change at runtime. 

We should probably as well print the operator config at the operator start.

internal/version/main.go
iblancasa: > Or should we make the version configurable in the CRD or the operator configuration, and set the docker image based on this (new) version field (see @rubenvp8510's comment)?

If we decide to get the version from the CRD or the operator configuration, the scenario is "closed" since the user will not be able to provide their own images, right? If we set the container image based on the version field, we can allow to set directly the images as the OpenTelemetry Operator does, right?

I was not sure about what approach to follow... so I decided to follow the first one. I think is something we can decide later if we see it is a use case we should cover. What do you think?

iblancasa:
@rubenvp8510
> I know may be not part of this PR but should we use the version as default image tag?

Good point! I'll do it.

main.go
iblancasa: > How is the tempo version propagated to the version command?
> 
> The tempo version can change and might change at runtime.
I prepared it to be configured at build time. As in the Jaeger Operator. From the @andreasgerstmayr comments, I'll change how the version is get.

> We should probably as well print the operator config at the operator start.

I did it [cmd/start/main.go](https://github.com/os-observability/tempo-operator/pull/184/files#diff-12831ff141f9eed16d3dc6aefbf0827971668f73f16bd3ba00ca26c69bbb9299).

andreasgerstmayr: > The tempo version can change and might change at runtime.

What is our support policy on that? I'd assume we only support the version of the default image as shipped with the default config, and if users use a custom image they need to make sure it's compatible (e.g. tempo v2.0 has breaking changes in the config file, I don't think we should support multiple tempo versions in the same operator version).

andreasgerstmayr (approved):

main.go
iblancasa: +1 @andreasgerstmayr

cmd/start/main.go
```diff
@@ -53,7 +55,18 @@ func start(c *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	setupLog.Info("starting manager")
```
frzifus, line 56: why did we remove this log line?

internal/version/main.go
frzifus: My understanding is that the operator in version x also only supports tempo un of version x. As well as the upgrade from x-1. As @andreasgerstmayr mentioned we can print the images from the spec.

cmd/start/main.go
iblancasa: If you check the diff, it was replaced by:
```go
	setupLog.Info("Starting Tempo Operator",
		"tempo-operator", v.OperatorVersion,
		"default-tempo-image", v.DefaultTempoImage,
		"default-tempo-query-image", v.DefaultTempoQueryImage,
		"build-date", v.BuildDate,
		"git-hash", v.GitHash,
		"go-version", v.Go,
		"go-arch", runtime.GOARCH,
		"go-os", runtime.GOOS,
	)
```

internal/version/main.go
iblancasa: I changed the code behavior before the comment to print the images from the spec so, we're good to go.

PR #190, opened 2023-02-02
Change the Makefile and Dockerfile to include the version metadata

iblancasa:

This PR is a continuation of #183.

What is included:
- Add the version, commit and build date medatada to the Makefile to build the `manager`
- Add the version, commit and build date medatada to the Docker to build the container image properly

What is not included (but will be included in future PRs):
- Change the Tempo version used by the operator
- Possible needed changes in the bundle
- Documentation


Dockerfile
```diff
@@ -16,8 +16,13 @@ COPY controllers/ controllers/
...
 COPY cmd/ cmd/
 
+ARG VERSION_PKG
+ARG COMMIT_SHA
+ARG OPERATOR_VERSION
+ARG VERSION_DATE
+
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-X ${VERSION_PKG}.buildDate=${VERSION_DATE} -X ${VERSION_PKG}.version=${OPERATOR_VERSION} -X ${VERSION_PKG}.commitSha=${COMMIT_SHA}" -a -o manager main.go
```
frzifus, line 25: Would it not make sense to call `make build` here instead?

versions.txt
```diff
@@ -0,0 +1,8 @@
+# this file contains the version of the Tempo components that will be used
+# by default with the Tempo Operator. This would usually be the latest
+# stable Tempo version. When you update this file, make sure to update the
+# the docs as well.
+tempo=1.5.0
```
pavolloffay, line 5: This info is as well defined in the https://github.com/os-observability/tempo-operator/blob/main/config/manager/controller_manager_config.yaml#L23

I would like to find an approach that minimizes duplications and makes it easy to bump versions. We should as well start a RELEASE doc and document how to bump versions for a release (might be a different PR).

pavolloffay (changes requested):

versions.txt
andreasgerstmayr: I think we don't need the tempo version there (`TEMPO_VERSION` is unused in the Makefile, because the version command now prints the default image).

About the operator version: can we use some git magic like `git describe --tags --always --dirty="-dev"` to generate the operator version?

Dockerfile
iblancasa: I'm following the same approach followed in OpenTelemetry Operator and Jaeger Operator. Should we change the 3 of them?

versions.txt
iblancasa: From the PR description:
> What is not included (but will be included in future PRs):
>
> * Change the Tempo version used by the operator
> * Possible needed changes in the bundle
> * Documentation

I'm following the same approach the Jaeger Operator and OpenTelemetry Operator are following. In the next PR I'll add the changes to modify the version of the used images and the documentation

Dockerfile
frzifus: I assume we should start here. But sure, why not.

iblancasa: I asked to create the same issue for OTEL and Jaeger Operators. TBH, I was thinking about this change but wanted to follow the same approach as in the other projects because I thought you would prefer consistency.

iblancasa:
Fixing CI...

Makefile
```diff
@@ -1,3 +1,18 @@
+# Current Operator version
+VERSION_DATE ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+VERSION_PKG ?= "github.com/os-observability/tempo-operator/internal/version"
+TEMPO_VERSION ?= "1.5.0"
+OPERATOR_VERSION ?= 0.0.1
```
iblancasa, line 5: Create an issue to replace this by `shell git describe --tags --always --dirty="-dev"` once we have the first tag. Otherwise, `make bundle` will fail always

Dockerfile
```diff
@@ -10,25 +10,17 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
-COPY main.go main.go
-COPY apis/ apis/
-COPY controllers/ controllers/
-COPY internal/ internal/
-COPY cmd/ cmd/
+COPY . .
```
frzifus, line 13: :+1:

frzifus (approved):

Makefile
```diff
@@ -115,11 +110,15 @@ vet: ## Run go vet against code.
...
+.PHONE: prepare-release
+prepare-release:
+	./hack/prepare-release.sh
+
 ##@ Build
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	go build -o bin/manager -ldflags ${LD_FLAGS} main.go
```
frzifus, line 121: Wait, i assume we will need:
```suggestion
	CGO_ENABLED=0 go build -o bin/manager -ldflags ${LD_FLAGS} main.go
```

Makefile
andreasgerstmayr: afaics it should also work in repos without tags, `--always` uses the last commit id as a fallback if the repo doesn't contain any tags. If I run it in my tempo-operator fork, it prints the last commit id.

iblancasa: Yes but the problem is when generating the bundle. It only accepts versioning using `x.y.z`. The commit SHA is not following that structure and the bundle generation fails.

andreasgerstmayr: Ah! Makes sense.
I didn't realize that there is a strict validation of the version number format.

Makefile
```diff
@@ -1,3 +1,18 @@
...
+OPERATOR_VERSION ?= 0.0.1
+COMMIT_SHA = "$(shell git rev-parse HEAD)"
+LD_FLAGS ?= "-X ${VERSION_PKG}.buildDate=${VERSION_DATE} -X ${VERSION_PKG}.version=${OPERATOR_VERSION} -X ${VERSION_PKG}.commitSha=${COMMIT_SHA}"
+ARCH ?= $(shell go env GOARCH)
+
+# Image URL to use all building/pushing image targets
+IMG_PREFIX ?= ghcr.io/${USER}/tempo-operator
+IMG_REPO ?= tempo-operator
+IMG ?= ${IMG_PREFIX}/${IMG_REPO}:${OPERATOR_VERSION}
+BUNDLE_IMG ?= ${IMG_PREFIX}/${IMG_REPO}-bundle:${OPERATOR_VERSION}
```
andreasgerstmayr, line 14: Should we prefix docker (version) tags with `v`? Currently `CATALOG_IMG` has the prefix, `BUNDLE_IMG` and `IMG` doesn't. imho either all of them should have the `v` prefix, or none.

hack/prepare-release.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "OPERATOR_VERSION isn't set. Skipping process."
+    exit 1
+fi
+
+echo "Updating the Tempo images version"
+sed -i "s~docker.io/grafana/tempo:.*~docker.io/grafana/tempo:${TEMPO_VERSION}~gi" config/manager/controller_manager_config.yaml
+sed -i "s~docker.io/grafana/tempo-query:.*~docker.io/grafana/tempo-query:${TEMPO_VERSION}~gi" config/manager/controller_manager_config.yaml
+
+echo "Updating the versions.txt file"
+sed -i "s~operator=.*~operator=${OPERATOR_VERSION}~gi" versions.txt
```
andreasgerstmayr, line 14: This file doesn't exist anymore.

andreasgerstmayr (approved):
lgtm, one reference to `versions.txt` should be removed.

Makefile
```diff
@@ -33,19 +41,8 @@ BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
-# IMAGE_TAG_BASE defines the docker.io namespace and part of the image name for remote images.
-# This variable is used to construct full image tags for bundle and catalog images.
-#
-# For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
-# grafana.com/tempo-operator-bundle:$VERSION and grafana.com/tempo-operator-catalog:$VERSION.
-IMAGE_TAG_BASE ?= grafana.com/tempo-operator
```
andreasgerstmayr, line 41: `CATALOG_IMG` still references this variable - can you update `CATALOG_IMG` to use `IMG_PREFIX` etc., like the `BUNDLE_IMG`?

andreasgerstmayr (changes requested):

andreasgerstmayr (approved):

frzifus:
ping @pavolloffay ^^

Makefile
```diff
@@ -115,11 +110,15 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet setup-envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONE: prepare-release
+prepare-release:
```
pavolloffay, line 114: can you please remove this for now (including the script) and do it in a separate PR that will introduce the release process?

pavolloffay (changes requested):

Makefile
iblancasa: Done

pavolloffay (approved):

PR #196, opened 2023-02-06
Add status of all components to CR status

rubenvp8510:


iblancasa (commented):
The E2E tests are failing. I think is not something related to the PR because I was not able to reproduce the issues locally... and it seems the deployments are not created.

How about rerunning?

internal/status/client.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/os-observability/tempo-operator/apis/tempo/v1alpha1"
+)
+
+type StatusClient interface {
+	GetPodsComponent(ctx context.Context, componentName string, stack v1alpha1.Microservices) (*corev1.PodList, error)
+	PatchStatus(ctx context.Context, changed, original *v1alpha1.Microservices) error
+}
+
+type StatusClientStub struct {
```
iblancasa, line 16: I'm curious: is there any special reason why this is suffixed as `Stub`? I'm asking because Stubs are a kind of Mocks (not exactly) and I would expect that something suffixed as `Stub` is related to testing

internal/status/status.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+
+	dockerparser "github.com/novln/docker-parser"
+
+	"github.com/os-observability/tempo-operator/apis/tempo/v1alpha1"
+)
+
+func Refresh(ctx context.Context, k StatusClient, tempo v1alpha1.Microservices, status *v1alpha1.MicroservicesStatus) (bool, error) {
+	tempoImage, err := dockerparser.Parse(tempo.Spec.Images.Tempo)
```
iblancasa, line 12: I know this comes from a previous PR and you are just refactoring but: should we show this information in the query too?

internal/status/client.go
rubenvp8510: It is only for tests, mm  🤔  this should not  be exposed outside the package

internal/status/status.go
rubenvp8510: for the query image? sure

internal/status/client.go
frzifus: Lets move the object into `client_test.go`

rubenvp8510: Done

controllers/tempo/microservices_controller.go
```diff
@@ -84,34 +72,60 @@ func (r *MicroservicesReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	var degraded *DegradedError
 	err := r.reconcileManifests(ctx, log, req, tempo)
+	if res, derr := r.handleDegradedError(ctx, tempo, err); derr != nil {
+		return res, derr
+	}
+
+	return r.refreshComponents(ctx, tempo)
+}
+
+func (r *MicroservicesReconciler) handleDegradedError(ctx context.Context, tempo v1alpha1.Microservices, err error) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
rubenvp8510: Good idea, the name is generic but really the function has an specific task, added comment.
Thanks!

internal/status/status.go
rubenvp8510: Done.

iblancasa (approved):
Great job!

frzifus (approved):
@rubenvp8510 lets create another issue to do the same for the tempo gateway

PR #208, opened 2023-02-10
fix: gateway tenant secert mount

frzifus:


pavolloffay (approved):

pavolloffay:
Assuming the unit tests will follow

PR #215, opened 2023-02-10
add basic gateway e2e tests

frzifus:
Signed-off-by: Benedikt Bongartz <bongartz@klimlive.de>

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -2,5 +2,5 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 testDirs:
   - ./tests/e2e/
-timeout: 300
-
+timeout: 600
+parallel: 2


tests/e2e/gateway/01-install.yaml
```diff
@@ -0,0 +1,53 @@
...
+  storageSize: 200M
+  tenants:
+    mode: static
+    authentication:
+      - tenantName: test-oidc
+        tenantId: test-oidc
+        oidc:
+          secret:
+            name: test-oidc
+          issuerURL: https://dex.klimlive.de/dex
```
iblancasa, line 34: Could we use something ownered by the team or something that can be started in the test environment?

tests/e2e/gateway/02-assert.yaml
```diff
@@ -0,0 +1,212 @@
...
+        secret:
+          defaultMode: 420
+          items:
+          - key: tenants.yaml
+            path: tenants.yaml
+          secretName: tempo-foo-gateway
+---
+apiVersion: v1
+data:
+  tenants.yaml: dGVuYW50czoKLSBuYW1lOiB0ZXN0LW9pZGMKICBpZDogdGVzdC1vaWRjCiAgb2lkYzoKICAgIGlzc3VlclVSTDogaHR0cHM6Ly9kZXgua2xpbWxpdmUuZGUvZGV4CiAgICAKICBvcGE6CiAgICBxdWVyeTogZGF0YS50ZW1wby5hbGxvdwogICAgcGF0aHM6CiAgICAtIC9ldGMvdGVtcG8tZ2F0ZXdheS9jbS9yYmFjLnlhbWwKICAgIC0gL2V0Yy90ZW1wby1nYXRld2F5L2NtL3RlbXBvLWdhdGV3YXkucmVnbw==
```
iblancasa, line 198: Do we want to check this value too?

kuttl-test.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 testDirs:
   - ./tests/e2e/
-timeout: 150
-
+timeout: 600
```
iblancasa, line 5: Maybe we can look for a better solution than increasing the timeout a lot.
Maybe, asserting the different deployments on the order they are started (one per KUTTL step).

iblancasa (changes requested):

tests/e2e/gateway/01-install.yaml
frzifus: We will do in future tests, here it actually doesnt matter. Since its not used.

kuttl-test.yaml
frzifus: It was more a simple test that give enough time to investigate why the compactor doesnt start. :D

tests/e2e/gateway/02-assert.yaml
iblancasa: Ok, you want to ensure the content is that one.

tests/e2e/gateway/01-install.yaml
iblancasa: So: since it is not used, you can remove it.

tests/e2e/gateway/02-assert.yaml
frzifus: if the value is the same, the content is the same.

tests/e2e/gateway/02-assert.yaml
```diff
@@ -0,0 +1,212 @@
+apiVersion: v1
+data:
+  rbac.yaml: "tenants:\n- name: test-oidc\n  id: test-oidc\n  oidc:\n    issuerURL: https://dex.klimlive.de/dex\n    \n  opa:\n    query: data.tempo.allow\n    paths:\n \   - /etc/tempo-gateway/cm/rbac.yaml\n    - /etc/tempo-gateway/cm/tempo-gateway.regodc"
```
andreasgerstmayr, line 3: Could you add this as multiline yaml instead of a string with encoded newlines?
https://github.com/kubernetes/kubernetes/issues/36222#issuecomment-1062121115

tests/e2e/gateway/01-install.yaml
andreasgerstmayr: Does it still work when `dex.klimlive.de` is replaced with `localhost`?

kuttl-test.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: kuttl.dev/v1beta1
...
+parallel: 2
```
andreasgerstmayr, line 6: Does the test also work without this setting?

tests/e2e/gateway/02-assert.yaml
iblancasa: +1

PR #228, opened 2023-02-13
gateway: support custom header to identify tenants

### LINKED ISSUE #224: Support enabling multi tenancy without gateway
Make sure the multitenancy can be enabled without gateway component (without authentication & authorization).

frzifus:
Closes #224

This commit extends the CR by a Multitenancy spec. I would like to hear your thoughts on it.

### DIFF
--- a/apis/tempo/v1alpha1/microservices_types.go
+++ b/apis/tempo/v1alpha1/microservices_types.go
@@ -166,6 +166,14 @@ const (
 // TenantsSpec defines the mode, authentication and authorization
 // configuration of the tempo gateway component.
 type TenantsSpec struct {
+	// Multitenancy support in tempo. Default is false.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Multitenancy enabled"
+	Multitenancy bool `json:"multitenancy"`
+
 	// Mode defines the multitenancy mode.
 	//
 	// +required
--- a/internal/manifests/gateway/build.go
+++ b/internal/manifests/gateway/build.go
@@ -106,11 +106,22 @@ const (
 	portPublic   = 8080
 )
 
+func getTenantHeaderFlag(t *tempov1alpha1.TenantsSpec) []string {
+	if t == nil {
+		return nil
+	}
+	if !t.Multitenancy {
+		return nil
+	}
+	return []string{fmt.Sprintf("--traces.tenant-header=%s", manifestutils.MultitenancyHeader)}
+}
+
 func deployment(params manifestutils.Params) *v1.Deployment {
 	tempo := params.Tempo
 	labels := manifestutils.ComponentLabels(tempoComponentName, tempo.Name)
 	annotations := manifestutils.CommonAnnotations(params.ConfigChecksum)
 	cfg := tempo.Spec.Components.Gateway
+	headerFlags := getTenantHeaderFlag(params.Tempo.Spec.Tenants)
 
 	return &v1.Deployment{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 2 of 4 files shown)


apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -162,9 +162,34 @@ const (
...
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Multitenancy enabled"
+	Enabled bool `json:"enabeld"`
+
+	// Header used to identify tenant.
+	// +optional
+	// +kubebuilder:example="X-Scope-OrgID"
+	Header string `json:"tempoQueryVersion,omitempty"`
```
pavolloffay, line 179: What is the use case for specifying the header? 

as far as I know tempo upstream supports only `X-Scope-OrgID`

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -162,9 +162,34 @@ const (
...
+// MultitenancySpec defines if gateway, qurier and db multitenancy support should
+// be enabled and which header is used to identify tenants.
+type MultitenancySpec struct {
+	// Enable multi tenancy support in tempo. Default is false.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Multitenancy enabled"
+	Enabled bool `json:"enabeld"`
```
pavolloffay, line 174: It probably needs more thinking what `enabled` flags are in the CR for multitenancy. There is already `spec.template.gateway.enabled` , we should make sure that the config is easy and does not require flipping multiple enabled flags for a single use-case.

pavolloffay (changes requested):

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -162,9 +162,34 @@ const (
 	Dynamic ModeType = "dynamic"
 )
 
+// MultitenancySpec defines if gateway, qurier and db multitenancy support should
```
iblancasa, line 165:
```suggestion
// MultitenancySpec defines if gateway, querier and db multitenancy support should
```

apis/tempo/v1alpha1/microservices_types.go
iblancasa, line 174:
```suggestion
	Enabled bool `json:"enabled"`
```

internal/manifests/gateway/build.go
```diff
@@ -106,11 +106,22 @@ const (
 	portPublic   = 8080
 )
 
+func getTenantHeaderFlag(t *tempov1alpha1.TenantsSpec) []string {
```
iblancasa, line 109: If this is a single flag... I would expect to get only one `string` returned, not a slice. WDYT?

iblancasa (changes requested):

apis/tempo/v1alpha1/microservices_types.go
frzifus: Iam not sure. Is it not configurable in tempo and jaeger?
https://github.com/jaegertracing/jaeger/blob/913ea321c9e1149871f7561bb69c497bd7316545/pkg/tenancy/flags.go#L35

nvm, ive no strong opinion on it. Its now replaced by a constant value.

internal/manifests/gateway/build.go
iblancasa: You're returning a `nil` or `1-element` slice. You can solve that with
```go
headerFlags := []string{getTenantHeaderFlag(params.Tempo.Spec.Tenants)}
```

frzifus: How would the function signature of `getTenantHeaderFlag` look like?

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: right, but it's not configurable in tempo. So without the gateway, the config does not make sense.

apis/tempo/v1alpha1/microservices_types.go
pavolloffay: Some form of validation/defaulting to make the configuration easier.

internal/manifests/gateway/build.go
pavolloffay: We should decide how to handle these `if-else` statements. 

The original idea was to use `mergo` and just merge manifests. I would rather stick to a single more readable approach.

frzifus: works for me too. So it would look like this:
```golang
patch := configureGatewayTenant(...)

if err := mergo.Merge(origin, patch, mergo.WithAppendSlice); err != nil {
	return nil, fmt.Errorf("failed to merge multi tenant settings into gateway container spec %w", err)
}
```

Issue #230, opened 2023-02-13
Rename default service account


pavolloffay:
The default SA is e.g. `tempo-simplest-serviceaccount`. The suffix seems redundant. The name `tempo-<instance-name>` should be enough and the same convention is used for other objects.

PR #231, opened 2023-02-13
Rename default service account

andreasgerstmayr:
Resolves: #230

### DIFF
--- a/internal/manifests/naming/naming.go
+++ b/internal/manifests/naming/naming.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-// Name returns component name.
+// Name returns the manifest name of a component.
 func Name(component string, instanceName string) string {
 	if component == "" {
 		return fmt.Sprintf("tempo-%s", instanceName)
@@ -14,5 +14,5 @@ func Name(component string, instanceName string) string {
 
 // DefaultServiceAccountName returns the name of the default tempo service account to use.
 func DefaultServiceAccountName(name string) string {
-	return Name("serviceaccount", name)
+	return Name("", name)
 }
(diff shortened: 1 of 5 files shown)


internal/manifests/naming/naming.go
```diff
@@ -14,5 +14,5 @@ func Name(component string, instanceName string) string {
 
 // DefaultServiceAccountName returns the name of the default tempo service account to use.
 func DefaultServiceAccountName(name string) string {
```
pavolloffay, line 16: should we rename this to `ServiceAccountName` ? 

I have also noticed that we are calling `Name` with empty component from other parts of the codebase. Myabe we can use that API directly and remove SA name function. WDYT?

andreasgerstmayr: > should we rename this to `ServiceAccountName` ?

I think it makes sense to have `Default` in the name, because it returns the _default_ service account. The service account name can be changed by the user in the CR, so one should not assume that the service account is always named after the return value of this function.

> I have also noticed that we are calling `Name` with empty component from other parts of the codebase. Myabe we can use that API directly and remove SA name function. WDYT?

I had exactly the same idea when sending the PR, but then I noticed we're also calling this function from the webhook, and I didn't want to encode the logic of "how we name the default SA" in the webhook, so I thought it's cleaner as it is.

otel-operator has its own naming package (https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/naming/main.go), which inspired me in the first place to create this tiny function. Maybe we should also add truncate functionality and check for invalid characters in the name in a follow-up PR?

iblancasa: > Maybe we should also add truncate functionality and check for invalid characters in the name in a follow-up PR?

I like this

iblancasa (approved):

internal/manifests/naming/naming.go
andreasgerstmayr: I've created #233 for sanitizing the generated names

rubenvp8510 (approved):

PR #236, opened 2023-02-16
Add tests to controller when TLS is enabled

rubenvp8510:


controllers/tempo/suite_test.go
```diff
@@ -26,6 +26,7 @@ func TestMain(m *testing.M) {
 		//	Paths: []string{filepath.Join("..", "config", "webhook")},
 		//},
 	}
+	testEnv.BinaryAssetsDirectory = "/home/rvargasp/.local/share/kubebuilder-envtest/k8s/1.24.2-linux-amd64/"
```
iblancasa, line 29: This should not be needed since `KUBEBUILDER_ASSETS` is set.

iblancasa (changes requested):
LGTM. Just a minor change

controllers/tempo/suite_test.go
rubenvp8510: Done, thanks!

andreasgerstmayr: It's required when you run the test from the "Testing" tab in VS Code. I'd also like this fixed somehow, but of course we don't want to hardcode this absolute path (including a version) :|

rubenvp8510: Yes  exactly, I tried to put in my `launch.json` the env variable `KUBEBUILDER_ASSETS` and it didn't work, so this was an easy "fix". :D

iblancasa (approved):

PR #252, opened 2023-02-23
TLS Profile support

rubenvp8510:
This PR adds TLS profile settings (`minVersion`, `ciphers`) to tempo configuration for the internal mTLS communication,

The profiles could be specified on the configuration features or in the case of _OpenShift_ it could be get from the cluster.

If the profile is invalid or is not specified (empty string) the default profile (Intermediate) will be applied, if `openshift.ClusterTLSPolicy` is enabled, it should get the profile from the cluster, if the cluster return a unknown profile this should return an error.

TLS Profiles: https://wiki.mozilla.org/Security/Server_Side_TLS


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
 	github.com/Azure/go-autorest/autorest v0.11.27 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.9.20 // indirect
-	github.com/openshift/api v0.0.0-20230213202419-42edf4f1d905
+	github.com/openshift/api v0.0.0-20220712151050-2647eb31dee7 // release-4.11
```
andreasgerstmayr, line 26: Should we use `release-4.14` instead? Or the last released version, `release-4.12`?

internal/tlsprofile/get.go
```diff
@@ -0,0 +1,38 @@
...
+	// If ClusterTLSPolicy is enabled get the policy from the cluster
+	if fg.OpenShift.ClusterTLSPolicy {
+		tlsProfileType, err = getTLSProfileFromCluster(ctx, c)
+	} else {
+		tlsProfileType, err = getTLSSecurityProfile(configv1alpha1.TLSProfileType(fg.TLSProfile))
+	}
+
+	if err != nil {
+		log.Error(err, "failed to get security profile. will use default tls profile.")
+		tlsProfileType = getDefaultTLSSecurityProfile()
```
andreasgerstmayr, line 28: I think for security configurations I would put tempo in a degraded/failed state instead of silently falling back to the defaults (I know there's a log, but I assume only very few people actually read the container logs).

iblancasa: +1

internal/tlsprofile/get.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	// Transform the policy type to concrete settings (cpyhers and minVersion).
```
iblancasa, line 31:
```suggestion
// Transform the policy type to concrete settings (ciphers and minVersion).
```

go.mod
rubenvp8510: may be yes :)

internal/tlsprofile/get.go
rubenvp8510: I propose this:

- if the profile is not specified on the gateway flags, we use the default, no degrade state.
- if the `openshift.ClusterTLSPolicy ` is enabled and failed to fetch the policy from the cluster, then we can think on return a degraded error. 

WDYT?

internal/tlsprofile/get.go
rubenvp8510: I was not able to pass the linter if I do this change

internal/tlsprofile/get.go
iblancasa: > I propose this:
> 
> * if the profile is not specified on the gateway flags, we use the default, no degrade state.
> * if the `openshift.ClusterTLSPolicy ` is enabled and failed to fetch the policy from the cluster, then we can think on return a degraded error.
> 
> WDYT?

I'm not sure about the "failing to fetch the policy from the cluster". I don't have a strong opinion but I think it can be a little bit difficult to detect.

andreasgerstmayr: > I propose this:
> 
>     * if the profile is not specified on the gateway flags, we use the default, no degrade state.
+1
>     * if the `openshift.ClusterTLSPolicy ` is enabled and failed to fetch the policy from the cluster, then we can think on return a degraded error.
Sounds good!

internal/tlsprofile/get.go
iblancasa: Why?

rubenvp8510: I think because the indent, this comment is an internal function comment, haaa I see the typo in the word, ok I will only include that part :D sorry.

rubenvp8510: done

internal/tlsprofile/get.go
rubenvp8510: Done with the behavior we agreed.

internal/tlsprofile/get.go
```diff
@@ -0,0 +1,45 @@
...
+	// If ClusterTLSPolicy is enabled get the policy from the cluster
+	if fg.OpenShift.ClusterTLSPolicy {
+		tlsProfileType, err = getTLSProfileFromCluster(ctx, c)
+		if err != nil {
+			returnedErr = ErrGetProfileFromCluster
+			tlsProfileType = getDefaultTLSSecurityProfile()
+		}
+	} else {
+		tlsProfileType, err = getTLSSecurityProfile(configv1alpha1.TLSProfileType(fg.TLSProfile))
+		if err != nil {
```
andreasgerstmayr, line 32: If I'm an admin and by mistake specify `featureGates.tlsProfile: modern` (note the lower case `modern`), then the operator will log an error but still use the default `Intermediate` profile?

I'd suggest to always return an error in this case, and at some other place in the code set the default value for the `featureGates.tlsProfile` setting. afaics currently we don't have a way to set default values for the `ProjectConfig` struct, I'd suggest we create a new function to set the default values for the `ProjectConfig` struct (similar to the Defaulter webhook) and call it in our `main.go` files directly after parsing the yaml file.

internal/tlsprofile/get.go
```diff
@@ -0,0 +1,45 @@
...
+// from the cluster, if the cluster return a unknow profile this should return an error.
+func Get(ctx context.Context, fg configv1alpha1.FeatureGates, c k8getter, log logr.Logger) (TLSProfileOptions, error) {
+	var tlsProfileType openshiftconfigv1.TLSSecurityProfile
+	var err error
+	var returnedErr error
+	// If ClusterTLSPolicy is enabled get the policy from the cluster
+	if fg.OpenShift.ClusterTLSPolicy {
+		tlsProfileType, err = getTLSProfileFromCluster(ctx, c)
+		if err != nil {
+			returnedErr = ErrGetProfileFromCluster
```
andreasgerstmayr, line 27: Could we return here immediately with `return TLSProfileOptions{}, ErrGetProfileFromCluster`?

internal/tlsprofile/get.go
rubenvp8510: I agree we should have a centralized way of setting defaults I think that could be done in a separate PR.

rubenvp8510: On the other hand what I don't like about returning an error here and allow other part of the code to handle it is that this function encapsulated all of things related to tlsprofile, if I do in other way I'll lose that encapsulation and I would need to add that logic to the controller or other piece of code.

rubenvp8510: I think what we should do in the future, (not in the PR) is to validate those inputs in some place, making sure we cannot allow invalid settings, and also set the defaults as you said. then I can get the value and use here. and we should not reach the error block for this funciton.

internal/tlsprofile/get.go
rubenvp8510: If we do that we won't be able to return the default tls options.

andreasgerstmayr: Maybe I'm misunderstanding something, but afaics the only call to `tlsprofile.Get` is in `microservices_controller.go`, and if this function returns an `ErrGetProfileFromCluster` error than the `tlsProfile` variable is unused?

internal/tlsprofile/get.go
andreasgerstmayr: > On the other hand what I don't like about returning an error here and allow other part of the code to handle it is that this function encapsulated all of things related to tlsprofile, if I do in other way I'll lose that encapsulation and I would need to add that logic to the controller or other piece of code.

I think any code must be able to return errors, I'd do it the same way as with the `ErrGetProfileFromCluster` error - return an error and let the error handling code in `microservices_controller.go` present the error to the user (via the status conditions).

> I think what we should do in the future, (not in the PR) is to validate those inputs in some place, making sure we cannot allow invalid settings (the operator can fail to start if that happens or log an error and set a default), and also set the defaults as you said. if that happens I can get the value and use here, we should not reach the error block for this function, because all values have been validated. Make sense?

Yep, setting defaults and validating the ProjectConfig in a separate PR sounds good to me.

andreasgerstmayr: I've created https://github.com/os-observability/tempo-operator/issues/266 to track the defaulter/validator for the ProjectConfig.

andreasgerstmayr (approved):

bundle/manifests/tempo-operator.clusterserviceversion.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    createdAt: "2023-03-01T18:24:59Z"
```
frzifus, line 17: oh - i assume we need something like this:
https://github.com/jaegertracing/jaeger-operator/blob/7e668d84b948b8366b46eaf5dfe0c0a849e943e4/hack/ignore-createdAt-bundle.sh

rubenvp8510: Yes

iblancasa (approved):

PR #255, opened 2023-02-23
Create Ingress or Route for Jaeger UI

andreasgerstmayr:
* if jaegerQuery is enabled, create an Ingress or Route object by default, depending on the OpenShift.Route feature gate
* show an error if Ingress or Route is enabled but jaegerQuery is disabled

TODO:
* Should we auto-detect OpenShift by querying ServerGroups, like jaeger-operator does? This would make the OpenShift.Route feature gate obsolete and would be a bit more user friendly imho


apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -41,6 +57,10 @@ type OpenShiftFeatureGates struct {
 	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
 	GatewayRoute bool `json:"gatewayRoute,omitempty"`
 
+	// Route enables creating OpenShift Route objects instead of Ingress objects by default.
+	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
+	Route bool `json:"route,omitempty"`
```
iblancasa, line 62: I would call this `OpenShiftRoute`. `Route` sounds too generic for me

apis/tempo/v1alpha1/microservices_types.go
```diff
@@ -367,24 +369,6 @@ type MicroservicesList struct {
 	Items           []Microservices `json:"items"`
 }
 
-// ImagesSpec defines the image for each container.
-type ImagesSpec struct {
```
iblancasa, line 371: Why did you need to change the location of this for creating the Ingress/OpenShift Routes?

config/rbac/role.yaml
```diff
@@ -42,6 +42,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - networking.k8s.io
```
iblancasa, line 46: Not directly related to this line but, I think something is missing for OpenShift Routes: 
Did you try to deploy the Tempo Operator in OpenShift? Maybe the `routes/custom-host` resource is missing.

[More information here](https://github.com/open-telemetry/opentelemetry-operator/pull/1337)

internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -53,13 +57,21 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
...
 	}
+
+	if tempo.Spec.Components.QueryFrontend.JaegerQuery.Ingress != nil && tempo.Spec.Components.QueryFrontend.JaegerQuery.Ingress.Enabled {
+		manifests = append(manifests, ingress(tempo))
+	}
+
+	if tempo.Spec.Components.QueryFrontend.JaegerQuery.Route != nil && tempo.Spec.Components.QueryFrontend.JaegerQuery.Route.Enabled {
+		manifests = append(manifests, route(tempo))
+	}
+
```
iblancasa, line 74: Just to understand: do we allow to enable `Ingress` and `OpenShift Routes` at the same time?
* If yes: why? I don't see the value of that and can cause some confusion
* If not: should that be validated in `validateQueryFrontend`?

iblancasa (changes requested):
Could we add a test for OpenShift Routes and Ingresses?

apis/tempo/v1alpha1/microservices_types.go
andreasgerstmayr: Currently `projectconfig_types.go` (`apis/config/v1alpha1`) has a dependency on `microservices_types.go` (`apis/tempo/v1alpha1`) to use `ImagesSpec`. For this PR I need access to the projectconfig in the webhook, so I need to import `apis/config/v1alpha1` from `apis/tempo/v1alpha1`, and then we have a circular dependency. Therefore I moved it to `apis/config/v1alpha1`.

internal/manifests/queryfrontend/query_frontend.go
andreasgerstmayr: +1, good idea to validate it in the webhook. I'll update the PR.

config/rbac/role.yaml
andreasgerstmayr: Yes, I tried it on OpenShift and it works (tested with crc). Pavol added the `route.openshift.io` RBAC as part of the gateway PRs.
I think `routes/custom-host` from the linked PR is only required if we set a custom hostname for the route (which is not yet possible in this PR).

Currently I've both a `ingress:` and `route:` field in the spec. Should I change it to how otel-operator has it with a single `ingress` field in the spec, and a `type` field which can toggle Ingress vs. Route?

andreasgerstmayr: My original idea why I created two separate fields for Route and Ingress that they may have different fields, for example some option is only available for Routes (e.g. custom TLS settings) and not for Ingress and vice-versa. Some are possible in both objects though, like hostname and annotations.

iblancasa: I see. Well, it totally makes sense in that case.

internal/manifests/queryfrontend/query_frontend.go
andreasgerstmayr: Configuring both at the same time is not possible anymore with the updated spec (`type: route|ingress`).

apis/tempo/v1alpha1/ingress_types.go
```diff
@@ -0,0 +1,36 @@
...
+
+type (
+	// IngressType represents how a service should be exposed (ingress vs route).
+	// +kubebuilder:validation:Enum=none;ingress;route
+	IngressType string
+)
+
+const (
+	// IngressTypeNone specifies that no ingress or route entry should be created.
+	IngressTypeNone IngressType = "none"
```
frzifus, line 11: Is none ever used? Seems its not set in the default webhook. 
```
IngressTypeUnknown IngressType = ""
```

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -41,6 +57,10 @@ type OpenShiftFeatureGates struct {
 	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
 	GatewayRoute bool `json:"gatewayRoute,omitempty"`
 
+	// OpenShiftRoute enables creating OpenShift Route objects instead of Ingress objects.
+	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
+	OpenShiftRoute bool `json:"openshiftRoute,omitempty"`
```
frzifus, line 62: Does it only belong to JaegerQuery or to all components that can benefit from ingress/routes?

If its for all components, it might overlaps with:
```golang
	// GatewayRoute enables creating an OpenShift Route for the TempoStack
	// gateway to expose the service to public internet access.
	// More details: https://docs.openshift.com/container-platform/latest/networking/understanding-networking.html
	GatewayRoute bool `json:"gatewayRoute,omitempty"`
```

apis/tempo/v1alpha1/ingress_types.go
andreasgerstmayr: It's required for users who explicitly want to disable the ingress. If users set ingress to empty, the defaulter webhook will enable the ingress again.

Basically I need a way to distinguish between "unset" (the webhook should enable ingress by default) and "disabled" in case the user wants to disable the ingress (then the webhook must not re-enable the ingress).

apis/config/v1alpha1/projectconfig_types.go
andreasgerstmayr: It's similar, yes.
Currently:
* `OpenShiftRoute` allows using Routes instead of Ingress, when a Route/Ingress is enabled in the CR.
* `GatewayRoute` *always* creates a Route for the gateway component for every tempo instance if this feature gate is enabled (with no option to enable/disable the Route in the CR).

Should there be an option in the spec to disable the gateway route per tempo instance? And, should there be an option to use Ingress instead of a Route for the gateway?

apis/tempo/v1alpha1/ingress_types.go
andreasgerstmayr: Alternatively we could decide to not enable the ingress by default. But I kind of like enabling it by default for UX reasons (I assume most people want it enabled).

andreasgerstmayr:
> Could we add a test for OpenShift Routes and Ingresses?

I added an assertion for the ingress in the smoketest-with-jaeger e2e.

I'm currently thinking how to make one for Route - the problem is we need to enable the OpenShiftRoutes feature gate in the test. I'm unsure if I should really update the operator configmap and restart the operator inside a kuttl test? Parallel kuttl tests will then fail of course, and imho the tests should not restart or update the operator's configuration.

Alternatively we could remove this feature gate and auto-detect the availability of openshift route. wdyt?

apis/tempo/v1alpha1/ingress_types.go
iblancasa: > Alternatively we could decide to not enable the ingress by default. But I kind of like enabling it by default for UX reasons (I assume most people want it enabled).

+1

andreasgerstmayr: > > Alternatively we could decide to not enable the ingress by default. But I kind of like enabling it by default for UX reasons (I assume most people want it enabled).
> 
> +1

+1 for not enabling by default, or +1 for enabling by default? :)

iblancasa:
> I'm currently thinking how to make one for Route - the problem is we need to enable the OpenShiftRoutes feature gate in the test. I'm unsure if I should really update the operator configmap and restart the operator inside a kuttl test? Parallel kuttl tests will then fail of course, and imho the tests should not restart or update the operator's configuration.

That is one of the reasons why I don't like running the tests in parallel.

I agree about the modification of the operator for the tests but... it is a common use case to test the different operator features. The last step of the test can be reverting back the changes done to the operator.

> Alternatively we could remove this feature gate and auto-detect the availability of openshift route. wdyt?

I don't have a strong opinion but, during some discussions, I understand the team prefers to use feature gates because it makes it easier to maintain the operators (because you enable or disable features instead of just detecting them). I don't have a strong opinion about this. @pavolloffay, @frzifus or @rubenvp8510 will provide a better answer than me about this topic.

apis/tempo/v1alpha1/ingress_types.go
frzifus: hm.. Not sure if i would like to expose something by default. ^^

apis/config/v1alpha1/projectconfig_types.go
frzifus: hm.. I think we should bring the two together. In a way that we can configure the object type and the default behavior per component (UI/Gateway).

apis/tempo/v1alpha1/ingress_types.go
andreasgerstmayr: You're right, maybe that enabled-by-default ingress comes as a surprise for some users, which is bad for security. I've removed the code to enable it by default, and now we can use `""` as ingress-is-disabled value.

apis/config/v1alpha1/projectconfig_types.go
andreasgerstmayr: I've created https://github.com/os-observability/tempo-operator/issues/265 to track this.

cmd/root.go
```diff
@@ -29,14 +29,14 @@ type RootConfigKey struct{}
...
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
 	utilruntime.Must(tempov1alpha1.AddToScheme(scheme))
-	utilruntime.Must(configtempov1alpha1.AddToScheme(scheme))
+	utilruntime.Must(configv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(routev1.Install(scheme))
 	utilruntime.Must(openshiftoperatorv1.Install(scheme))
 	utilruntime.Must(configv1.Install(scheme))
```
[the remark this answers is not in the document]
andreasgerstmayr: They're already registered: `utilruntime.Must(routev1.Install(scheme))` (afaics it was part of some gateway PR).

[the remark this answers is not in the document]
frzifus: ups :D

frzifus (approved):

iblancasa (approved):

rubenvp8510 (approved):

Issue #271, opened 2023-03-03
Add support for Azure Storage


iblancasa:
Support Azure Storage as storage for Tempo Operator. [Loki Operator](https://loki-operator.dev/docs/object_storage.md/#azure) has support for it.

PR #276, opened 2023-03-07
Add the storage type to the CRD

iblancasa:
## Why this PR is needed
As a first step to introducing new storages (#271)

## What changes are introduced in this PR
* Followed the same approach as the [Loki Operator](https://loki-operator.dev/docs/object_storage.md/) for the CRD
* Fixed documentation

Following PRs will add support for the different storages but I'm trying to do smaller PRs.


README.md
```diff
@@ -44,7 +44,9 @@ metadata:
   name: simplest
 spec:
   storage:
-    secret: minio-test
+    secret:
+      name: minio-test
+      type: s3
```
andreasgerstmayr, line 49: What's the reasoning behind the storage type being a property of `secret` instead of `storage`?

iblancasa: I had the same question. I copied the schema from the [`Loki Operator`](https://loki-operator.dev/docs/object_storage.md/) (for consistency) and I think is because the fields from the secret are different depending on the used storage.

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -431,10 +458,9 @@ type ObjectStorageSpec struct {
 	// Name of a secret in the same namespace as the tempo TempoStack custom resource.
 	//
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MinLength=1
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Object Storage Secret"
-	Secret string `json:"secret"`
-	// Don't forget to update storageSecretField in tempostacks_controller.go if this field name changes.
+	Secret ObjectStorageSecretSpec `json:"secret"`
+	// Don't forget to update storageSecretField in tempostack_controller.go if this field name changes.
```
andreasgerstmayr, line 463: Can you update the `storageSecretField` constant in `tempostack_controller.go`? It specifies which field should be indexed (and it'll reconcile all TempoStacks which use the same referenced secret). I'll create an additional e2e test for this (separate PR), afaics the current e2e test should have failed but apparently did not fail.

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -417,6 +417,33 @@ type SearchSpec struct {
...
+)
+
+// ObjectStorageSecretSpec is a secret reference containing name only, no namespace.
+type ObjectStorageSecretSpec struct {
+	// Type of object storage that should be used
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:s3"},displayName="Object Storage Secret Type"
+	Type ObjectStorageSecretType `json:"type"`
```
pavolloffay, line 437: What is the default? 

What happens if the type is not specified? Does the CR create fail or is it defaulted to `s3`?

pavolloffay (approved):

Issue #259, opened 2023-02-27
Prune no-longer-used manifests


andreasgerstmayr:
We need a generic way to prune unused manifests.

Example:
1. A user enables the Ingress in the CR
2. The tempo-operator creates an Ingress object
3. The user disables Ingress in the CR
4. The Ingress object is still present because the operator does not delete it from the cluster

I think a generic way would be to run the following in the Reconcile function:
* get a list of all resources with ownerRef = current operator instance
* delete all resources which are not returned by the `BuildAll()` function (`manifests.go`)

PR #280, opened 2023-03-08
Prune unmanaged resources

andreasgerstmayr:
Prune objects which exist in the cluster but are not managed by the operator anymore. For example, when the Jaeger Query Ingress is enabled and later disabled, the Ingress object should be removed from the cluster.

Resolves: #259

### DIFF
--- a/controllers/tempo/tempostack_controller_test.go
+++ b/controllers/tempo/tempostack_controller_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


andreasgerstmayr:
Could we increase the covscan limit a bit, or do we really want to test every error path? :neutral_face:
(I'd need to find a way to fake a kubernetes API error when deleting an object for example)

controllers/tempo/tempostack_controller_test.go
```diff
@@ -419,3 +421,74 @@ func TestTLSEnable(t *testing.T) {
...
+	}
+	req := ctrl.Request{
+		NamespacedName: nsn,
+	}
+	reconcileResult, err := reconciler.Reconcile(context.Background(), req)
+	require.NoError(t, err)
+	assert.Equal(t, false, reconcileResult.Requeue)
+
+	// Verify Ingress is created
+	ingressNsn := types.NamespacedName{Name: "tempo-prune-ingress-test-query-frontend", Namespace: "default"}
```
iblancasa, line 473: Maybe, you should do the same for routes (just to increase the coverage)

iblancasa (approved):

controllers/tempo/tempostack_controller_test.go
andreasgerstmayr: The tempostack_controller tests use a Kubernetes control plane (`KUBEBUILDER_ASSETS` env var in the Makefile).
When I create a unit test with a Route object I get a panic when calling `obj.GetName()` on the Route object.
I already expected that it won't work, but I'm surprised that it panics :smiling_face_with_tear:

andreasgerstmayr: So afaics we can only test Routes in e2e tests, when the OpenShift router is installed.

Issue #266, opened 2023-03-01
Create Defaulter and Validator for ProjectConfig


andreasgerstmayr:
We should create a `Defaulter` and `Validator` for the `ProjectConfig` (the tempo-operator configuration), in a similar fashion how we set defaults and validate settings of the CR.

Setting the defaults and validating the configuration should happen during the startup of the operator.
If the validation fails, the operator should exit immediately.

Ref. discussion in https://github.com/os-observability/tempo-operator/pull/252#discussion_r1121924195

pavolloffay:
>  (the tempo-operator configuration), in a similar fashion how we set defaults and validate settings of the CR.

Do I assume correctly that the impl will not sue the webhooks?

andreasgerstmayr:
> > (the tempo-operator configuration), in a similar fashion how we set defaults and validate settings of the CR.
> 
> Do I assume correctly that the impl will not sue the webhooks?

Yes, this issue is about the operator configuration, which afaics never get validated through webhooks (the config needs to be valid before the webhook is started).

PR is in #281

PR #281, opened 2023-03-09
Set default TLSProfile and validate ProjectConfig

andreasgerstmayr:
Resolves: #266

We don't need a dedicated Default() function, as we can just set default values in `root.go` when we create an instance of `ProjectConfig`

### DIFF
--- a/apis/config/v1alpha1/projectconfig_types.go
+++ b/apis/config/v1alpha1/projectconfig_types.go
@@ -1,6 +1,8 @@
 package v1alpha1
 
 import (
+	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cfg "sigs.k8s.io/controller-runtime/pkg/config/v1alpha1"
 )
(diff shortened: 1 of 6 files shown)


cmd/testdata/tlsprofile_invalid.yaml
```diff
@@ -0,0 +1,2 @@
+featureGates:
```
iblancasa, line 1: Instead of creating files for this, how about doing something like `internal/manifests/config/build_test.go`? We can have the content of the file there, write it before the test to a `tmp` folder, and read the content from there.

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -144,3 +146,17 @@ type ProjectConfig struct {
 func init() {
 	SchemeBuilder.Register(&ProjectConfig{})
 }
+
+// Validate validates the controller configuration (ProjectConfig).
+func (c *ProjectConfig) Validate() error {
```
iblancasa, line 151: Could we validate more stuff? Some suggestions:
* `.metrics.bindAddress` is a valid address
* `.health.healthProbeBindAddress` is a valida ddress
* `.webhook.port` is a valid port
* `.images[]` follow valid container name image schema

andreasgerstmayr: I've added validation for images (part of our custom/additional operator settings). The other settings come from the controller-runtime project, and are validated there (the operator binary will print an error message on startup if they're invalid).

iblancasa (approved):

PR #283, opened 2023-03-09
Add support for Azure Storage

### LINKED ISSUE #271: Add support for Azure Storage

iblancasa:
Fixes #271 

Merge after #276


controllers/tempo/tempostack_controller.go
```diff
@@ -241,10 +241,10 @@ func (r *TempoStackReconciler) getStorageConfig(ctx context.Context, tempo v1alp
 		return nil, fmt.Errorf("invalid storage secret: %s", strings.Join(msgs, ", "))
 	}
 
-	return &manifestutils.StorageParams{S3: &manifestutils.S3{
-		Endpoint: string(storageSecret.Data["endpoint"]),
-		Bucket:   string(storageSecret.Data["bucket"]),
-	}}, nil
+	return &manifestutils.StorageParams{
+		S3:           getS3Params(storageSecret),
+		AzureStorage: getAzureParams(storageSecret),
```
andreasgerstmayr, line 246: The storageSecret can be either a S3 secret or an Azure secret, but never both (or am I missing something?).
I'd conditionally either extract the S3 or the Azure params from the secret.

internal/manifests/manifestutils/params.go
```diff
@@ -17,7 +17,15 @@ type Params struct {
 
 // StorageParams holds storage configuration.
 type StorageParams struct {
-	S3 *S3
+	S3           *S3
+	AzureStorage *AzureStorage
+}
+
+// AzureStorage for Azure Storage.
+type AzureStorage struct {
```
andreasgerstmayr, line 25: Would it make sense to consolidate all Azure storage params in a single struct?
* `manifestutils.AzureStorage`
* `manifests.AzureStorage`
* `config.azureStorage` (options.go)
* `config.AzureStorage` (configmap.go)

controllers/tempo/tempostack_controller.go
iblancasa: I though the same but followed the Loki Operator approach.
The code takes care later of just getting the params associated with the desired storage. I don't have a strong opinion about this. Maybe, I would implement it in the same way you pointed but I maybe it is good to keep the same approach as Loki. I don't  know

internal/manifests/manifestutils/params.go
iblancasa: I had a similar feeling but followed the same approach as for S3 in this project and as Loki Operator for all the storages.

controllers/tempo/tempostack_controller.go
andreasgerstmayr: I took a peek and afaics loki does have a switch based on the type when extracting the storage params from the secret:
https://github.com/grafana/loki/blob/f125682802a556e6b759fce13e1a0c930b57f90f/operator/internal/handlers/internal/storage/secrets.go#L11-L36

iblancasa: You're totally right. I'll do the change

internal/manifests/manifestutils/params.go
iblancasa: That's true. I'm not sure why in Tempo the approach was different.

Since is a refactor that should be done for S3 too... what do you think if I create a follow up task to address it there?

andreasgerstmayr: Yep, follow-up task sounds good, thanks!

apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -171,43 +170,6 @@ func (v *validator) validateServiceAccount(ctx context.Context, tempo TempoStack
 	return allErrs
 }
 
-// ValidateStorageSecret validates the object storage secret required for tempo.
-func ValidateStorageSecret(tempo TempoStack, storageSecret corev1.Secret) field.ErrorList {
```
frzifus, line 175: why is this moved out of the webhook file, when its used only there?

iblancasa: It is used in:
* `controllers/tempo/tempostack_controller.go`
* `apis/tempo/v1alpha1/tempostack_webhook.go`

And I move it to `apis/tempo/v1alpha1/storage.go` with all the helper functions I created to do the validation. I think it totally makes sense to avoid making the `apis/tempo/v1alpha1/tempostack_webhook.go` bigger and, at least for me, is more intuitive in this way.

Issue #229, opened 2023-02-13
rename TempoComponentsSpec entry in crd


frzifus:
https://github.com/os-observability/tempo-operator/blob/542c7fb47a6d986f4d077c676ba681f5601ae583/apis/tempo/v1alpha1/microservices_types.go#L71

For me its a bit confusing that `Components` after json marshalling ends up as `template`. Whats your thoughts on that?

andreasgerstmayr:
Yes that's weird, I'd recommend renaming it in json to `components`.

PR #288, opened 2023-03-10
Rename template field in CRD to components

andreasgerstmayr:
Resolves: #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ spec:
       limits:
         memory: 2Gi
         cpu: 2000m
-  template:
+  components:
     queryFrontend:
       jaegerQuery:
         enabled: true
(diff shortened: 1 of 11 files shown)


frzifus (approved):
Oh yes and no :X

iblancasa (approved):

iblancasa (approved):

andreasgerstmayr:
Because it's quite a breaking change, maybe let's wait a bit for additional reviews.

rubenvp8510 (approved):

pavolloffay (approved):

pavolloffay (changes requested):
https://github.com/grafana/loki/blob/1b2f3e0cb2ebe2f012ebaa0731a0623ad6c537e2/operator/apis/loki/v1/lokistack_types.go#L725

We should prefer to be aligned with Loki CR and therefore rename the struct and not the CR field. 


Is there any strong argument to rename the CR field?

andreasgerstmayr:
> Is there any strong argument to rename the CR field?

I'm fine either way, the only thing bothering me was that the struct attribute name (Components) was different to the json field name (template).

PR #290, opened 2023-03-10
generate cmd: fix reading from stdin, add test

andreasgerstmayr:


cmd/generate/main_test.go
```diff
@@ -63,3 +65,46 @@ metadata:
...
+
+	c.SetArgs([]string{"generate"})
+	_, err := c.ExecuteC()
+	require.NoError(t, err)
+
+	require.Contains(t, out.String(), `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
```
iblancasa, line 101: Maybe: should we fix this in the future?  A `null` timestamp sounds weird for me

andreasgerstmayr: Done

iblancasa (commented):
LGTM. It seems the CI needs some fixes

andreasgerstmayr:
CI passes, codecov just complains that the file output path isn't tested. unit tests writing to the filesystem can be problematic, therefore imho it's still good to merge.

andreasgerstmayr:
codecov is now happy :)

iblancasa (commented):
Good job!
It seems the coverage is not happy again

frzifus (approved):
Iam fine with ignoring codecov here :baby_chick:

PR #299, opened 2023-03-13
Support trace reports across operands

### LINKED ISSUE #291: Operator exposes traces about the Operand
Not sure if it makes sense to have this configurable or simply send all traces to localhost.

See:
https://github.com/grafana/tempo/blob/d5e734c40216ad0db00c7c54cbe1ef685d90681a/docs/sources/tempo/operations/monitoring.md#traces
https://github.com/observatorium/api/blob/37022fddcd512d5b104c8ecfa22c4215d8cd7ff2/main.go#L241

frzifus:
Closes #291 

docs: https://github.com/frzifus/tempo-operator/blob/3e3c9a7e08eb4df5819c45336395c467cdf394b7/docs/tempostack/observability_tracing.md


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -10,6 +10,8 @@ import (
 
 // TempoStackSpec defines the desired state of TempoStack.
 type TempoStackSpec struct {
+	// TracingEnabled - TODO: find a better home.
+	TracingEnabled bool
```
frzifus, line 14: What would you recommend where that should live?

pavolloffay:
Did you test this end2end with injected OTELcol sidecar? We should as well document how the entire tracing should be setup (e.g. via the OTEL sidecar etc.)

apis/tempo/v1alpha1/tempostack_types.go
pavolloffay: I woulder if we should create a sub spec for internal observability. Is there any metrics related config we should set. Or perhaps sampling?

internal/manifests/manifestutils/tracing.go
```diff
@@ -0,0 +1,41 @@
...
+				Name:  "JAEGER_AGENT_PORT",
+				Value: "6831",
+			},
+			{
+				Name:  "JAEGER_SAMPLER_TYPE",
+				Value: "const",
+			},
+			{
+				Name:  "JAEGER_SAMPLER_PARAM",
+				Value: "1",
```
iblancasa, line 28: Should this be configurable somehow? I'm not saying to do it for this PR, it is more because I'm curious about

internal/manifests/querier/querier.go
```diff
@@ -79,6 +87,24 @@ func deployment(params manifestutils.Params) (*v1.Deployment, error) {
 							Name:  "tempo",
 							Image: tempo.Spec.Images.Tempo,
 							Args:  []string{"-target=querier", "-config.file=/conf/tempo.yaml"},
+							Env: []corev1.EnvVar{
```
iblancasa, line 90: These values are the same than in `PatchTracingJaegerEnv` right? Would it make sense to place them in a common location?

iblancasa, line 90: These values are the same than in `PatchTracingJaegerEnv` right? Would it make sense to place them in a common location?

apis/tempo/v1alpha1/tempostack_types.go
frzifus: y, could defently make sense. The thing is, what its used for. If it is only used for debugging purposes, I would hide it in the crd under `spec.internal.debugging`. Otherwise maybe something like `spec.internal.traceconfig` that keeps maybe things like `sampling` and `endpoint`.
Or a more "granular" option would be to place trace configs next to the component configurations. wdyt?

internal/manifests/manifestutils/tracing.go
frzifus: It depends how we want to use it. ^^

https://github.com/os-observability/tempo-operator/pull/299#discussion_r1135425230

internal/manifests/querier/querier.go
frzifus: ups - I think this can be removed since its gets patched here: https://github.com/os-observability/tempo-operator/pull/299/files#diff-7ecc0111bcea034df8f745984c2cc8629813d58f0e4162fa18a43a971272ec84R25

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -85,6 +85,25 @@ type TempoStackSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tenants Configuration"
 	Tenants *TenantsSpec `json:"tenants,omitempty"`
+
+	Observability ObservabilitySpec `json:"observability,omitempty"`
```
frzifus, line 89: wdyt about this @iblancasa @pavolloffay ?

iblancasa: I think it's OK. It's generic enough to add new fields later if needed.

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -85,6 +85,47 @@ type TempoStackSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Observability"
+	Observability ObservabilitySpec `json:"observability,omitempty"`
+}
+
+// ObservabilitySpec defines how telemetry data gets handled.
+type ObservabilitySpec struct {
+	// Operands defines how operand telemetry data gets handled.
+	Operands ObservabilityOperandsSpec `json:"operands,omitempty"`
```
pavolloffay, line 100: Why is the operand nesting needed? The CR is implicitly a config for operands only.

frzifus: true :X

rubenvp8510 (approved):

docs/tempostack/observability_tracing.md
```diff
@@ -0,0 +1,170 @@
...
+draft: false
+images: []
+menu:
+  docs:
+    parent: "tempostack"
+weight: 100
+toc: true
+---
+
+All tempo components as well as the tempo gateway ([observatorium-api](https://github.com/observatorium/api)) support the export of traces in `jaeger_compact` format.
```
frzifus, line 16:
```suggestion
All tempo components as well as the tempo gateway ([observatorium-api](https://github.com/observatorium/api)) support the export of traces in `thrift_compact` format.
```

PR #311, opened 2023-03-15
Avoid creation of query frontend route if gateway is enabled

rubenvp8510:
@frzifus

### DIFF
--- a/internal/manifests/queryfrontend/query_frontend.go
+++ b/internal/manifests/queryfrontend/query_frontend.go
@@ -64,12 +64,14 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 		manifests = append(manifests, s)
 	}
 
-	//exhaustive:ignore
-	switch tempo.Spec.Template.QueryFrontend.JaegerQuery.Ingress.Type {
-	case v1alpha1.IngressTypeIngress:
-		manifests = append(manifests, ingress(tempo))
-	case v1alpha1.IngressTypeRoute:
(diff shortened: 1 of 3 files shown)


iblancasa (approved):

internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -64,12 +64,14 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 		manifests = append(manifests, s)
 	}
 
-	//exhaustive:ignore
-	switch tempo.Spec.Template.QueryFrontend.JaegerQuery.Ingress.Type {
-	case v1alpha1.IngressTypeIngress:
-		manifests = append(manifests, ingress(tempo))
-	case v1alpha1.IngressTypeRoute:
-		manifests = append(manifests, route(tempo))
+	if !tempo.Spec.Template.Gateway.Enabled {
```
andreasgerstmayr, line 67: In case the user mistakenly enables both, now the ingress/route silently doesn't get created and the user might be left wondering why. Could you add a validation in the webhook to explicitly tell the user that he/she cannot enable ingress/route and gateway at the same time?

rubenvp8510: Good point. I'll add a validation.

rubenvp8510: Added

frzifus (approved):

andreasgerstmayr (approved):

PR #317, opened 2023-03-16
Do not fail CI if codecov step fails

andreasgerstmayr:
Unfortunately codecov is flaky lately, and it shouldn't affect the result of the unit test action.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -32,7 +32,7 @@ jobs:
           files: cover.out
           flags: unittests
           name: codecov-tempo-operator
-          fail_ci_if_error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
           files: cover.out
           flags: unittests
           name: codecov-tempo-operator
-          fail_ci_if_error: true
+          fail_ci_if_error: false
```
pavolloffay, line 35: https://github.com/codecov/codecov-action/issues/926

pavolloffay (approved):

frzifus (approved):

Issue #247, opened 2023-02-21
Reload tenant configuration 


pavolloffay:
Gateway needs to be restarted if tenant configuration changes

iblancasa:
Oh yes. I though the issue was automatically closed

PR #325, opened 2023-03-17
Reload tenant configuration

iblancasa:
Solves #247


andreasgerstmayr (approved):
Looks great!
I think an additional e2e test would be a good idea, possibly in the same test as #215.

iblancasa:
> Looks great! I think an additional e2e test would be a good idea, possibly in the same test as #215.

Thanks for your feedback @andreasgerstmayr!

@frzifus should I address this in a different PR or will you cover it in #329?

iblancasa:
From this PR I found the `gateway.go` file is not covered by the unit tests at all.

frzifus (approved):

frzifus:
> @frzifus should I address this in a different PR or will you cover it in https://github.com/os-observability/tempo-operator/pull/329?

Its not covered, but we can easily add it afterwards.

PR #326, opened 2023-03-19
support internal mTLS between gateway and upstream services

rubenvp8510:


frzifus (commented):
n1 - we might should add some tests. ^^

PR #329, opened 2023-03-20
add basic gateway e2e tests

frzifus:


tests/e2e/gateway/00-install.yaml
```diff
@@ -0,0 +1,157 @@
...
+  clientID: test
+  clientSecret: ZXhhbXBsZS1hcHAtc2VjcmV0
+type: Opaque
+---
+apiVersion: tempo.grafana.com/v1alpha1
+kind: TempoStack
+metadata:
+  name: foo
+spec:
+  images:
```
iblancasa, line 113: Do we need to add the images? It will force us to modify this file for each release, right?

tests/e2e/gateway/00-install.yaml
```diff
@@ -0,0 +1,157 @@
+---
```
iblancasa, line 1: I think it would be a good idea if we split this file into two steps:
1. Create the Secrets, ConfigMaps, and other needed stuff and start the app...
2. Create the TempoStack instance

This will help the debuggability of the test. Why? KUTTL has not the best output. If we are checking the creation of the different configurations + the deployment of the Tempo component, it can be difficult to find if what failed is something related to the "needed ingredients" or something the Tempo Operator should have created.

iblancasa (commented):
Do we plan to have follow-up tickets to test if the stored data per tenant work properly?

tests/e2e/gateway/00-install.yaml
frzifus: It doesnt work with the 2.0.1 images. Next time we update the file, we can remove it.

tests/e2e/gateway/00-install.yaml
frzifus: sure, i will do.

frzifus:
> Do we plan to have follow-up tickets to test if the stored data per tenant work properly?

yes

tests/e2e/gateway/00-assert.yaml
pavolloffay, line 1: It does not seem necessary to assert on all these fields for dex. It would be more useful to assert on if dex is actually running.

tests/e2e/gateway/02-install.yaml
```diff
@@ -0,0 +1,51 @@
+---
+apiVersion: tempo.grafana.com/v1alpha1
+kind: TempoStack
+metadata:
+  name: foo
```
pavolloffay, line 5: could this be changed to something better?

PR #344, opened 2023-03-25
Initial release process

rubenvp8510:


frzifus (commented):
Should we also add something like this [changelog.yaml](https://github.com/open-telemetry/opentelemetry-operator/blob/c4f8ffe8e221d9e0f7ce134de96f7aee02c3678d/.github/workflows/changelog.yaml) workflow?

rubenvp8510:
> Should we also add something like this [changelog.yaml](https://github.com/open-telemetry/opentelemetry-operator/blob/c4f8ffe8e221d9e0f7ce134de96f7aee02c3678d/.github/workflows/changelog.yaml) workflow?

I would like to, but first want to have this in place and then I will enforce the changelog stuff in a separate PR, WDYT?

test
frzifus, line 1: one last thing, do we need this file? :baby_chick:

rubenvp8510: not really :D

frzifus (approved):

.ci/create-release-github.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+OPERATOR_VERSION=$(git describe --tags)
```
pavolloffay, line 3: There is really no need for this script and it can be directly done in the GH action step.

PR #346, opened 2023-03-27
Fix OpenShift e2e gateway test to use cabundle

pavolloffay:
Fix after https://github.com/os-observability/tempo-operator/pull/343 was merged

### DIFF
--- a/tests/e2e-openshift/gateway/00-assert.yaml
+++ b/tests/e2e-openshift/gateway/00-assert.yaml
@@ -39,6 +39,25 @@ metadata:
       name: simplest
 ---
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/component: gateway
+    app.kubernetes.io/created-by: tempo-controller
+    app.kubernetes.io/instance: simplest
+    app.kubernetes.io/managed-by: tempo-controller
+    app.kubernetes.io/name: tempo
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+  name: tempo-simplest-gateway-cabundle
+  ownerReferences:
+    - apiVersion: tempo.grafana.com/v1alpha1
+      blockOwnerDeletion: true
+      controller: true
+      kind: TempoStack
+      name: simplest
+---
+apiVersion: v1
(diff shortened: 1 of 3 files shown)


tests/e2e-openshift/gateway/00-assert.yaml
```diff
@@ -147,9 +167,12 @@ spec:
           - --log.level=info
           - --tls.server.cert-file=/etc/tempo-gateway/serving-certs/tls.crt
           - --tls.server.key-file=/etc/tempo-gateway/serving-certs/tls.key
-          image: quay.io/observatorium/api:main-2023-02-09-v0.1.2-329-g1ff4f11
+          - --tls.internal.server.cert-file=/etc/tempo-gateway/serving-certs/tls.crt
+          - --tls.internal.server.key-file=/etc/tempo-gateway/serving-certs/tls.key
+          - --tls.healthchecks.server-ca-file=/etc/tempo-gateway/cabundle/service-ca.crt
+          - --tls.healthchecks.server-name=tempo-simplest-gateway.kuttl-ocp-gateway.svc.cluster.local
```
pavolloffay, line 173: because of this I have used the `--namespace` with kuttl. It seems the simplest solution for now.

tests/e2e-openshift/gateway/00-install.yaml
```diff
@@ -1,3 +1,8 @@
+# The namespace is auto-deleted by kuttl after the test run.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kuttl-ocp-gateway
```
iblancasa, line 5: Is this change needed? I think it would be enough setting the `namespace` field in `kuttl-test-openshift.yaml` (as you did). From the [KUTTL documentation](https://kuttl.dev/docs/testing/reference.html#testsuite):
> The namespace to use for tests. This namespace will be created if it does not exist and removed if it was created (unless skipDelete is set). If no namespace is set, one will be auto-generated.

pavolloffay: it was not created when I run it locally

iblancasa (approved):

PR #347, opened 2023-03-27
Run e2e OCP tests as part of release instructions

pavolloffay:


RELEASE.md
```diff
@@ -3,12 +3,13 @@
...
 
 1. Change the images tags to corresponding versions on `config/manager/controller_manager_config.yaml` the operator is usually aligned with the tempo versions. 
-3. Run `make bundle USER=os-observability OPERATOR_VERSION=x.y.z`, where `x.y.z` is the version that will be released.
-5. Add the changes to the changelog, see Generating the changelog secction.
-6. Send a PR with the changes
-7. Once the changes above are merged and available in `main`, tag it with the desired version, prefixed with `v`: `vx.y.z`
-8. The GitHub Workflow will take it from here, creating a GitHub release with the generated artifacts (manifests) and publishing the images
-9. After the release, generate a new OLM bundle (`make bundle`) and create two PRs against the `Community Operators repositories`:
+1. Run `make bundle USER=os-observability OPERATOR_VERSION=x.y.z`, where `x.y.z` is the version that will be released
+1. Run OpenShift tests locally against an OpenShift cluster `make e2e-openshift`. A locally installed [CRC](https://github.com/crc-org/crc) cluster can be used for testing.
```
iblancasa, line 7: The tests for Kubernetes are not mentioned because they are runned directly as part of the CI.
For the OpenShift tests, I think we should mention you first need to build the operator image and push it somewhere. Otherwise, the cluster will not be able to pull the image.
Also, the `deploy` target should be run (or the tests will not pass because the operator is not installed).

iblancasa: > That is somehow implicit, but I can add a note for it in the text.

Well, not totally: If the instructions say "run `make e2e-openshift`" I would expect that command to do everything for me.

PR #353, opened 2023-03-29
Run scorecard tests on CI

pavolloffay:


.github/workflows/scorecard.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: "install kuttl"
+        run: ./hack/install/install-kuttl.sh
+
+      - name: "start kind"
+        env:
+          KUBE_VERSION: ${{ matrix.kube-version }}
+        run: kind create cluster --config kind-$KUBE_VERSION.yaml
+
+      - name: "wait until cluster is ready"
+        run:  kubectl wait --timeout=5m --for=condition=available deployment/coredns -n kube-system
```
iblancasa, line 42: Maybe we should add this step to the `start-kind` make rule and use that rule for our CI workflows.
What do you think? We can do that as part of another PR

iblancasa (approved):

.github/workflows/scorecard.yaml
pavolloffay: feel free to put it there. I had some issues with readiness before so I put it here.

PR #356, opened 2023-03-29
Fix Makefile bundle rule

rubenvp8510:
Fixes https://github.com/os-observability/tempo-operator/issues/355

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,16 +211,10 @@ generate-bundle: operator-sdk manifests kustomize ## Generate bundle manifests a
 	$(OPERATOR_SDK) bundle validate $(BUNDLE_DIR)
 	./hack/ignore-createdAt-bundle.sh
 
-.PHONY: bundle-openshift
-bundle-openshift:
-	BUNDLE_VARIANT=openshift $(MAKE) generate-bundle
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -211,11 +211,16 @@ generate-bundle: operator-sdk manifests kustomize ## Generate bundle manifests a
 	$(OPERATOR_SDK) bundle validate $(BUNDLE_DIR)
 	./hack/ignore-createdAt-bundle.sh
 
+.PHONY: bundle-openshift
+bundle-openshift:
```
pavolloffay, line 215: I will ask again :) Why do we need to have two targets? 

I would advocate for a single target and make sure the two bundles are always generated and up-to-date simultaneously (also the makefile will be simpler).

pavolloffay (approved):

Makefile
rubenvp8510: Done! 1 target to rule them all!

pavolloffay: :P

pavolloffay (approved):

iblancasa (approved):

frzifus (approved):

PR #375, opened 2023-04-05
Make API package more readable

pavolloffay:
Just refactoring, moving the `TempoStack` and list as last types in the file


frzifus (approved):
What does it change?

pavolloffay:
Nothing moves the core structs to the end of the file

Issue #333, opened 2023-03-21
Enable/disable service monitors per TempoStack instance


andreasgerstmayr:
ServiceMonitors should be enabled/disabled per TempoStack instance (in the `observability` field in the spec)

Follow-up of #324
Requires CRD changes of #299

PR #381, opened 2023-04-12
Enable ServiceMonitors per TempoStack instance

andreasgerstmayr:
In the `.spec.observability.metrics.enabled` field of the TempoStack spec.

Resolves: #333


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -104,6 +111,16 @@ type ObservabilitySpec struct {
...
 }
 
+// MetricsConfigSpec defines a metrics config.
+type MetricsConfigSpec struct {
+	// Enabled specifies if a ServiceMonitor should be created for each Tempo component.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enabled"
+	Enabled bool `json:"enabled,omitempty"`
```
pavolloffay, line 121: I wonder if we should be more explicit in the name - e.g. `CreateServiceMonitor` . 

Also do we know if Tempo will stick with p8s metrics? Are there any plans to move to OTLP metrics? cc) @frzifus

apis/tempo/v1alpha1/tempostack_webhook_test.go
```diff
@@ -877,18 +877,59 @@ func TestValidateTenantConfigs(t *testing.T) {
 }
 
 func TestValidatorObservabilityTracingConfig(t *testing.T) {
-	tracingBase := field.NewPath("spec").Child("template").Child("observability").Child("tracing")
+	observabilityBase := field.NewPath("spec").Child("observability")
+	metricsBase := observabilityBase.Child("tracing")
```
iblancasa, line 881: Wouldn't it be?
```suggestion
	metricsBase := observabilityBase.Child("metrics")
```

apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -267,7 +267,17 @@ func (v *validator) validateGateway(tempo TempoStack) field.ErrorList {
 }
 
 func (v *validator) validateObservability(tempo TempoStack) field.ErrorList {
-	tracingBase := field.NewPath("spec").Child("template").Child("observability").Child("tracing")
+	observabilityBase := field.NewPath("spec").Child("observability")
+	metricsBase := observabilityBase.Child("tracing")
```
iblancasa, line 271: Maybe
```suggestion
	metricsBase := observabilityBase.Child("metrics")
```

apis/tempo/v1alpha1/tempostack_types.go
frzifus: y, i guess being a bit more explicit would make sense.

> Are there any plans to move to OTLP metrics?

Not that i would know of. I couldnt find a ticket and there is no otel metric instrumentation in their code.

andreasgerstmayr: Ok, I've updated the field name and displayName.

frzifus (approved):

PR #382, opened 2023-04-13
Generate one certificate per component

rubenvp8510:
Instead of having grpc/http certificates this PR generate one TLS cert/key per service, this is more aligned with the way we are creating the services which expose different ports on the same service.


andreasgerstmayr (approved):

internal/manifests/manifestutils/service.go
```diff
@@ -117,65 +111,3 @@ func ConfigureGRPCServicePKI(tempoStackName string, component string, podSpec *c
...
-
-	for _, i := range containersSlice {
-		if i >= nContainers {
-			continue
-		}
-		if err := mergo.Merge(&podSpec.Containers[i], secretContainerSpec, mergo.WithAppendSlice); err != nil {
-			return kverrors.Wrap(err, "failed to merge container")
-		}
-
-		if podSpec.Containers[i].ReadinessProbe == nil {
```
iblancasa, line 172: Why is this no longer needed?

PR #387, opened 2023-04-14
Add mTLS to gateway

rubenvp8510:
Fixes https://github.com/os-observability/tempo-operator/issues/240


internal/tlsprofile/options.go
```diff
@@ -17,3 +21,19 @@ type TLSProfileOptions struct {
...
+	case string(openshiftconfigv1.VersionTLS10):
+		return "1.0"
+	case string(openshiftconfigv1.VersionTLS11):
+		return "1.1"
+	case string(openshiftconfigv1.VersionTLS12):
+		return "1.2"
+	case string(openshiftconfigv1.VersionTLS13):
+		return "1.3"
+	default:
+		return ""
```
iblancasa, line 37: Would it make sense to return an error saying "unsupported version" or something?

internal/manifests/gateway/gateway.go
```diff
@@ -58,6 +58,25 @@ func BuildGateway(params manifestutils.Params) ([]client.Object, error) {
...
+		}
+		err := manifestutils.ConfigureServicePKI(params.Tempo.Name, manifestutils.DistributorComponentName, &dep.Spec.Template.Spec)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if params.Gates.HTTPEncryption {
+		dep.Spec.Template.Spec.Containers[0].Args = append(dep.Spec.Template.Spec.Containers[0].Args,
+			fmt.Sprintf("--traces.tls.key-file=%s/tls.key", manifestutils.TempoServerTLSDir()),
```
andreasgerstmayr, line 74: I like how for the query-frontend you modified the deployment in the `deployment()` function, I'd suggest doing the same here also, i.e. moving the code which modifies the deployment spec to the `deployment()` function.

In case someone is modifying the deployment spec, the chances are higher to also modify this code (if needed) if they're in the same function and not in different functions.

andreasgerstmayr (approved):

andreasgerstmayr (commented):
The generate e2e test fails currently with `Error: error building manifests: invalid TLS version`

andreasgerstmayr (approved):

iblancasa (approved):

Issue #385, opened 2023-04-13
Support Kubernetes 1.27


pavolloffay:

pavolloffay:
This will resolve #365 as well

pavolloffay:
done

Issue #365, opened 2023-04-03
Support kubernetes 1.26


pavolloffay:
The operator should be tested against Kubernetes 1.26

iblancasa:
I think we can close this issue.

pavolloffay:
done

PR #395, opened 2023-04-19
Add support for Kubernetes 1.27

iblancasa:
Solves #385 and  #365.

* Upgrade KiND to 0.18.0
* Upgrade Kubernetes used versions
* Added support for Kubernetes 1.26
* Added support for Kubernetes 1.27


.github/workflows/e2e.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
        # should be compatible with them.
        kube-version:
        - "1.21"
-       - "1.25"
+       - "1.27"
```
frzifus, line 24: Shall we do it on 2 steps? That the tests for 1.26 did run at least once? Just for completion :)

iblancasa: From [this comment](https://github.com/os-observability/tempo-operator/issues/385#issuecomment-1508041834), I supposed is not needed. Also, our thinking is that: if it works with the older and newer versions... we assume it works with the versions between.

kind-1.24.yaml
```diff
@@ -2,4 +2,18 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
-    image: kindest/node:v1.24.4@sha256:adfaebada924a26c2c9308edd53c6e33b3d4e453782c0063dc0028bdebaddf98
+    image: kindest/node:v1.24.12@sha256:1e12918b8bc3d4253bc08f640a231bb0d3b2c5a9b28aa3f2ca1aee93e1e8db16
+  kubeadmConfigPatches:
```
andreasgerstmayr, line 6: Are these new config options required because of the upgrade from `v1.24.4` to `v1.24.12`?

iblancasa: It seems we forget to add those options to 1.24. They should be there from the beginning

andreasgerstmayr (approved):

PR #401, opened 2023-04-28
Add a smoketest for tempo + opentelemetry-collector + multitenancy on OpenShift

andreasgerstmayr:
* Update trace verification of smoketest-with-jaeger because the wget executable may not be present in downstream images
* Move test setup commands of kuttl-test-openshift to prepare-e2e-openshift task, because these commands should not be run in all cases (e.g. if tempo is deployed via OLM or the desired config is already present)
* Update kuttl to fix test cleanup (fixed in kuttl 0.14)
* Improve error message when jaeger query ingress and gateway are enabled at the same time


Makefile
```diff
@@ -307,18 +307,28 @@ deploy-minio:
...
+prepare-e2e: kuttl start-kind cert-manager install-openshift-routes deploy-minio set-test-image-vars set-test-operator-config build docker-build load-image-operator deploy
+
 .PHONY: e2e
 e2e:
 	$(KUTTL) test
 
-# end-to-tests
+# OpenShift end-to-tests
+.PHONY: prepare-e2e-openshift
+prepare-e2e-openshift: deploy-minio
```
pavolloffay, line 320: any reason why this was moved form the `kuttl-test-openshift.yaml` to the makefile?

pavolloffay: The original idea was that the kuttl suite will configure the operator.

pavolloffay (commented):
I would merge the smoke test with the gateway test on OCP. I don't see much value of having both tests.

Makefile
andreasgerstmayr: > any reason why this was moved form the `kuttl-test-openshift.yaml` to the makefile?

Yes: when testing downstream products, I do not want to touch the config, and the namespace (tempo-operator-system) doesn't match.

Makefile
```diff
@@ -307,18 +307,28 @@ deploy-minio:
...
+	kubectl apply -f ./bundle/openshift/manifests/tempo-operator-manager-config_v1_configmap.yaml -n tempo-operator-system
+	kubectl rollout restart deployment/tempo-operator-controller-manager -n tempo-operator-system
+	kubectl rollout status deployment/tempo-operator-controller-manager -n tempo-operator-system --timeout=30s
+
 .PHONY: e2e-openshift
 e2e-openshift:
+	-kubectl delete namespace kuttl-e2e-openshift
```
iblancasa, line 327: Does the `-` character have an special meaning here?

tests/e2e-openshift/otel-multitenancy-smoketest/02-generate-traces.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: batch/v1
```
iblancasa, line 1: For this step, I would set a high nunmber for the duration and just assert the next step (verify the traces were received)

tests/e2e/smoketest-with-jaeger/02-generate-traces.yaml
```diff
@@ -7,7 +7,7 @@ spec:
     spec:
       containers:
         - name: tracegen
-          image: ghcr.io/open-telemetry/opentelemetry-collector-contrib/tracegen:latest
+          image: ghcr.io/open-telemetry/opentelemetry-collector-contrib/tracegen:v0.70.0
```
iblancasa, line 10: I know is not part of this task but, could you replace this by `telemetrygen`?

kuttl-test-openshift.yaml
```diff
@@ -1,11 +1,14 @@
...
 testDirs:
   - ./tests/e2e-openshift/
 timeout: 150
-namespace: kuttl-ocp-gateway
-commands:
-  - script: |
-      kubectl apply -f ./bundle/openshift/manifests/tempo-operator-manager-config_v1_configmap.yaml -n tempo-operator-system
-      kubectl rollout restart deployment/tempo-operator-controller-manager -n tempo-operator-system
-      kubectl rollout status deployment/tempo-operator-controller-manager -n tempo-operator-system --timeout=30s
+namespace: kuttl-e2e-openshift  # some resources depend on a known namespace (ClusterRoleBinding, URLs)
```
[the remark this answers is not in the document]
andreasgerstmayr: Great idea! In this case I have to explicitly configure the namespace for every manifest, and kuttl will still generate a random namespace for it, but won't actually use it.
If we use different (but static/known) namespaces for each test, we can also run them in parallel. Awesome :)

tests/e2e/smoketest-with-jaeger/02-generate-traces.yaml
andreasgerstmayr: sure, done.

tests/e2e-openshift/otel-multitenancy-smoketest/02-generate-traces.yaml
andreasgerstmayr: Do you suggest removing `02-assert.yaml`? I think it's good to know if the traces got generated successfully, and only after that verify that the traces are there. Afaics the Job has retry built in (via the Kubernetes Job object), but in the normal case there shouldn't be any retry, if we wait until the previous job (generate traces) was successful.

Makefile
andreasgerstmayr: imho we have currently two suites:
* e2e: must work on *all platforms*, using either the *community or OpenShift* tempo operator configuration, deployed via *manifests or OLM*
* e2e-openshift: must work on OpenShift, using the OpenShift tempo operator configuration, deployed via OLM

Now, if we want to test the OpenShift suite on e.g. minikube, somewhere _external_ to the test the missing pieces need to be configured: route-controller, cert-manager, opentelemetry-operator, and (probably) replacing the operator config. imho this could/should be in the makefile, but in "fine grained" tasks instead of catch-all `prepare-e2e*` tasks, for example we don't want to configure cert-manager if the operator was deployed via OLM, etc.

Issue #307, opened 2023-03-14
Alerting for operands


pavolloffay:
https://sdk.operatorframework.io/docs/overview/operator-capabilities/#alerting-and-events

AhmedGrati:
Hey @pavolloffay, I'm interested to tackle this issue, but it seems a bit blurry for me. Can you clarify it more, please? Maybe some subtasks, previous implementations, etc... can be more than helpful for me. Thanks 🙏

PR #410, opened 2023-05-16
create alerts rules for tempo stack

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/307


iblancasa (commented):
Please, could you add the link to the issue this PR solves? Thanks!

iblancasa (approved):

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -119,6 +119,13 @@ type MetricsConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create ServiceMonitors for Tempo components"
 	CreateServiceMonitors bool `json:"createServiceMonitors,omitempty"`
+
+	// EnableAlerts specifies if Alerts should be created for Tempo components.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create ServiceMonitors for Tempo components"
+	EnableAlerts bool `json:"enableAlerts,omitempty"`
```
andreasgerstmayr, line 128: Now we have `CreateServiceMonitors` and `EnableAlerts` fields, which is inconsistent.
Imho we should either rename `CreateServiceMonitors` -> `EnableServiceMonitors` or `EnableAlerts` -> `CreateAlerts`, wdyt?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -119,6 +119,13 @@ type MetricsConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create ServiceMonitors for Tempo components"
 	CreateServiceMonitors bool `json:"createServiceMonitors,omitempty"`
+
+	// EnableAlerts specifies if Alerts should be created for Tempo components.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create ServiceMonitors for Tempo components"
```
andreasgerstmayr, line 127:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Create Alerts for Tempo components"
```

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: yep, sounds good

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -155,6 +155,10 @@ type FeatureGates struct {
 	// ServiceMonitor defines whether the ServiceMonitor CRD exists in the cluster.
 	// This CRD is part of prometheus-operator.
 	ServiceMonitor bool `json:"serviceMonitor,omitempty"`
+
+	// EnableAlerts defines if  PrometheusRules will be created for alerting.
+	// This CRD is part of prometheus-operator.
+	EnableAlerts bool `json:"enableAlerts,omitempty"`
```
andreasgerstmayr, line 161: To be consistent with the `ServiceMonitor` feature gate above, maybe call this one `PrometheusRule`?

Calling it `EnableAlerts` sounds like it will actually enable alerts, however this feature gate only specifies if the `PrometheusRule` CRD exists in the cluster, and does not create any alerts by itself.

However, thinking more of it, if the `ServiceMonitor` CR exists then also the `PrometheusRule` CR exists. Maybe we should merge both flags into one `PrometheusOperator` flag or something? What are your thoughts?

rubenvp8510: Not sure about merge it in only one, but on the other side, I cannot think in a use case where you want to enable serviceMonitors and disable Rules.

andreasgerstmayr: I'd only merge this feature gate, not the flag in the CRD. So this featuregate basically only says "prometheus operator is available or not", and in the TempoStack CR users can enable or disable metrics and alerts independently (well, if you enable alerts but disable metrics than the alerts won' work...hmm).

andreasgerstmayr (approved):
Looks great, thanks for the changes!

Issue #411, opened 2023-05-16
Support a managed state like loki does


frzifus:
Its useful to debug things:
https://github.com/grafana/loki/blob/01f0ded7fcb57e3a7b26ffc1e8e3abf04a403825/operator/apis/loki/v1/lokistack_types.go#L13

cc @andreasgerstmayr

pavolloffay:
Things to consider:
* upgrade
* webhooks

PR #412, opened 2023-05-16
Do not set Tempo log level to debug

andreasgerstmayr:
The log level shouldn't be debug by default.
For debugging sessions, users should be able to customize the tempo config directly by using the unmanaged mode proposed in #411


frzifus (approved):
Does it make sense to set it to `error` level to not depend on the implementation details?

andreasgerstmayr:
> Does it make sense to set it to `error` level to not depend on the implementation details?

idk, the default log level is `info`, which is imho an appropriate default log level.

pavolloffay (approved):

PR #413, opened 2023-05-17
Add OLM targets to Makefile

andreasgerstmayr:
Running `make olm/install-operator` on a fresh minikube/kind cluster will install OLM and setup tempo-operator with a single command. Useful for locally debugging OLM related issues.


iblancasa:
Would it make sense to use this way of installing the operator for the E2E tests?

iblancasa (approved):

frzifus:
> Would it make sense to use this way of installing the operator for the E2E tests?

Yes, i think so.

frzifus (approved):

PR #420, opened 2023-05-24
Improve version handling

### LINKED ISSUE #400: Improve tempo version detection
Currently we set the `.status.tempoVersion` to the tag of the tempo image:
https://github.com/os-observability/tempo-operator/blob/17e668461afda9d0c118e26cc41f3604594aaefb/internal/status/status.go#L34

That's a problem for downstream products which replace the image and for example use a pinned image:

```
$ k -n kuttl-e2e-openshift get tempostack                                                     
NAME       AGE     TEMPO VERSION
simplest   4h50m   sha256:2ec3a...
```

andreasgerstmayr:
* store tempo and tempo-query versions in binary
* replace the image locations with versions in the `Version` struct
* set `.status.tempoVersion` based on the tempo version in the binary, not the docker tag (as the docker tag can point to a hash, e.g. in downstream products) 

Resolves #400


internal/version/main.go
```diff
@@ -6,57 +6,57 @@ import (
...
+	promversion.Version = operatorVersion
 	promversion.BuildDate = buildDate
-	promversion.Revision = commitSha
+	promversion.Revision = revision
 	metrics.Registry.MustRegister(promversion.NewCollector("tempooperator"))
 }
 
 // Get returns the Version object with the relevant information.
-func Get(config v1alpha1.ProjectConfig) Version {
+func Get() Version {
```
iblancasa, line 39: Why were the images removed?

iblancasa: Ok, I just saw you're getting them from a different place

iblancasa (approved):

Issue #296, opened 2023-03-13
Support upgrades of the instance


pavolloffay:
```[tasklist]
### Tasks
- [x] Create (empty for now)  upgrade procedure
- [x] Write upgrade test that simulates upgrade on OCP (using OLM and bundle)
```

andreasgerstmayr:
@pavolloffay how should the upgrade test work?

Something like
* install the currently released 0.1.0 from operatorhub
* build a bundle from the current git commit
* upgrade via OLM and make sure the images in the deployments are replaced?

pavolloffay:
👍🏼  LGTM

PR #424, opened 2023-05-26
Implement operator upgrade

andreasgerstmayr:
Resolves: #296

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -28,35 +28,3 @@ OPERATOR_VERSION=x.y.z make chlog-update
 ```
 
 This will delete all entries (other than the template) in the `.chloggen` directory and create a populated Changelog.md entry.
-
-## Test OLM Upgrade
-Install latest released version:
-```
-IMG_PREFIX=ghcr.io/grafana/tempo-operator OPERATOR_VERSION=old_xyz OPERATOR_NAMESPACE=openshift-operators make olm-deploy
-```
-
-Build and push operator and bundle image to a container registry:
-```
-IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z BUNDLE_VARIANT=openshift make bundle docker-build docker-push bundle-build bundle-push
-```
-
-Upgrade to new version:
-```
-IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z OPERATOR_NAMESPACE=openshift-operators make olm-upgrade
-```
-
-## Test Upgrade using Manifests
-Install latest released version:
-```
-kubectl apply -f https://github.com/os-observability/tempo-operator/releases/latest/download/tempo-operator.yaml
-```
-
-Build and push operator to a container registry:
-```
-IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z make docker-build docker-push release-artifacts
-```
-
-Upgrade to new version:
-```
-kubectl apply -f dist/tempo-operator.yaml
-```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -44,3 +44,19 @@ Upgrade to new version:
 ```
 IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z OPERATOR_NAMESPACE=openshift-operators make olm-upgrade
 ```
+
+## Test Upgrade using Manifests
+Install latest released version:
+```
+kubectl apply -f https://github.com/os-observability/tempo-operator/releases/latest/download/tempo-operator.yaml
+```
+
+Build and push operator to a container registry:
+```
+IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z make docker-build docker-push release-artifacts
+```
+
+Upgrade to new version:
+```
+kubectl apply -f dist/tempo-operator.yaml
+```
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -228,10 +228,6 @@ type TempoStackStatus struct {
 	// +optional
 	TempoVersion string `json:"tempoVersion,omitempty"`
 
-	// Version of the Tempo Query component used.
-	// +optional
-	TempoQueryVersion string `json:"tempoQueryVersion,omitempty"`
-
 	// Components provides summary of all Tempo pod status grouped
 	// per component.
 	//
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -220,12 +220,18 @@ type ComponentStatus struct {
 
 // TempoStackStatus defines the observed state of TempoStack.
 type TempoStackStatus struct {
+	// Version of the Tempo Operator.
+	// +optional
+	OperatorVersion string `json:"operatorVersion,omitempty"`
+
 	// Version of the managed Tempo instance.
 	// +optional
 	TempoVersion string `json:"tempoVersion,omitempty"`
+
 	// Version of the Tempo Query component used.
 	// +optional
 	TempoQueryVersion string `json:"tempoQueryVersion,omitempty"`
+
 	// Components provides summary of all Tempo pod status grouped
 	// per component.
 	//
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -228,6 +228,10 @@ type TempoStackStatus struct {
 	// +optional
 	TempoVersion string `json:"tempoVersion,omitempty"`
 
+	// DEPRECATED. Version of the Tempo Query component used.
+	// +optional
+	TempoQueryVersion string `json:"tempoQueryVersion,omitempty"`
+
 	// Components provides summary of all Tempo pod status grouped
 	// per component.
 	//
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -660,7 +660,8 @@ type RetentionConfig struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
-//+kubebuilder:printcolumn:name="Tempo version",type="string",JSONPath=".status.tempoVersion",description="Tempo Version"
+//+kubebuilder:printcolumn:name="Tempo Version",type="string",JSONPath=".status.tempoVersion",description="Tempo Version"
+//+kubebuilder:printcolumn:name="Management",type="string",JSONPath=".spec.managementState",description="Management State"
 
 // TempoStack is the spec for Tempo deployments.
 //
--- a/apis/tempo/v1alpha1/tempostack_webhook.go
+++ b/apis/tempo/v1alpha1/tempostack_webhook.go
@@ -71,9 +71,7 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 	if r.Labels == nil {
 		r.Labels = map[string]string{}
 	}
-	if r.Spec.ManagementState == ManagementStateUnmanaged {
-		delete(r.Labels, "app.kubernetes.io/managed-by")
-	} else {
+	if r.Labels["app.kubernetes.io/managed-by"] == "" {
 		r.Labels["app.kubernetes.io/managed-by"] = "tempo-operator"
 	}
 	r.Labels["tempo.grafana.com/distribution"] = d.ctrlConfig.Distribution
--- a/apis/tempo/v1alpha1/tempostack_webhook.go
+++ b/apis/tempo/v1alpha1/tempostack_webhook.go
@@ -71,7 +71,9 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 	if r.Labels == nil {
 		r.Labels = map[string]string{}
 	}
-	if r.Labels["app.kubernetes.io/managed-by"] == "" {
+	if r.Spec.ManagementState == ManagementStateUnmanaged {
+		delete(r.Labels, "app.kubernetes.io/managed-by")
+	} else {
 		r.Labels["app.kubernetes.io/managed-by"] = "tempo-operator"
 	}
 	r.Labels["tempo.grafana.com/distribution"] = d.ctrlConfig.Distribution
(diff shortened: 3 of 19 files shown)


andreasgerstmayr:
Currently working on an OLM test as described in https://github.com/grafana/tempo-operator/issues/296

andreasgerstmayr:
> Currently working on an OLM test as described in #296

Getting OLM to upgrade from a custom catalog is more involved than I thought :smiling_face_with_tear:, and I'd like to also parameterize so it also works also for downstream products (i.e. arbitrary catalogs and bundles), so I'll split the automated OLM upgrade test out to #438. This PR is ready for review (and also contains a few unit tests).

internal/upgrade/suite_test.go
```diff
@@ -0,0 +1,54 @@
...
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var testScheme *runtime.Scheme = scheme.Scheme
+
+func TestMain(m *testing.M) {
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
+		//WebhookInstallOptions: envtest.WebhookInstallOptions{
+		//	Paths: []string{filepath.Join("..", "config", "webhook")},
+		//},
```
iblancasa, line 27: Maybe we can remove this

internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Spec: v1alpha1.TempoStackSpec{
+			Images: configv1alpha1.ImagesSpec{
+				Tempo:           "docker.io/grafana/tempo:0.0.0",
+				TempoQuery:      "docker.io/grafana/tempo-query:0.0.0",
+				TempoGateway:    "quay.io/observatorium/api:0.0.0",
+				TempoGatewayOpa: "quay.io/observatorium/opa-openshift:0.0.0",
+			},
+			Storage: v1alpha1.ObjectStorageSpec{
+				Secret: v1alpha1.ObjectStorageSecretSpec{
+					Type: "s3",
```
iblancasa, line 39: Should this be `ObjectStorageSecretS3`?

iblancasa (commented):
We found one problem in the OpenTelemetry Operator related to the upgrade mechanism and I would like to avoid it here.

The problem was that it was assumed the "Defaulter Webhook" was run not just for new instances but for upgraded instances. This was not happening and, one value that should have been set by the "Defaulter Webhook" was not set. When the upgrade of the OTEL Operator was happening, causing a crash. Is there any way we can ensure we're running the "Defaulter Webhook" before the upgrade mechanism to ensure everything works as expected?

andreasgerstmayr:
> Part of this task was to implement as well e2e upgrade test using OLM. Is it going to be added?

I did split the OLM upgrade task into a new task because of size and complexity, here is the PR: https://github.com/grafana/tempo-operator/pull/441

apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -73,7 +73,9 @@ func (d *Defaulter) Default(_ context.Context, obj runtime.Object) error {
 	if r.Labels == nil {
 		r.Labels = map[string]string{}
 	}
-	if r.Labels["app.kubernetes.io/managed-by"] == "" {
+	if r.Spec.ManagementState == ManagementStateUnmanaged {
+		delete(r.Labels, "app.kubernetes.io/managed-by")
```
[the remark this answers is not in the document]
andreasgerstmayr: Yes, this change removes the `managed-by` label if the operand is unmanged.
The upgrade process filters for instances with the `managed-by` label and doesn't upgrade unmanaged instances.

internal/upgrade/upgrade.go
```diff
@@ -0,0 +1,142 @@
...
+			tempo = *upgraded
+		}
+	}
+
+	// update all tempo images to the new default images on every upgrade
+	updateTempoStackImages(u, &tempo)
+	// at the end of the upgrade process, the CR is up to date with the current running versions
+	tempo.Status.OperatorVersion = u.Version.OperatorVersion
+	tempo.Status.TempoVersion = u.Version.TempoVersion
+	tempo.Status.TempoQueryVersion = u.Version.TempoQueryVersion
```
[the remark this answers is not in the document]
andreasgerstmayr: done

internal/upgrade/suite_test.go
andreasgerstmayr: done

internal/upgrade/upgrade_test.go
andreasgerstmayr: :+1:, done

andreasgerstmayr:
> We found one problem in the OpenTelemetry Operator related to the upgrade mechanism and I would like to avoid it here.
> 
> The problem was that it was assumed the "Defaulter Webhook" was run not just for new instances but for upgraded instances. This was not happening and, one value that should have been set by the "Defaulter Webhook" was not set. When the upgrade of the OTEL Operator was happening, causing a crash.

The defaulter webhook was supposed to, but did not run before or after the upgrade?

> Is there any way we can ensure we're running the "Defaulter Webhook" before the upgrade mechanism to ensure everything works as expected?

Under what circumstances was the default webhook not run before the upgrade? Was this another (unrelated) issue, causing the webhook to be not installed or something? It's supposed to run always when changing the CR, and at the point when the operand is about to get upgraded, the CR should already have the default values.

I did test the current code, and the defaulter webhook runs after the upgrade process - the upgrade patches the CR in-place, and at the time when it patches the CR in the cluster, the defaulter and validating webhook run automatically.

andreasgerstmayr:
> > Under what circumstances was the default webhook not run before the upgrade? Was this another (unrelated) issue, causing the webhook to be not installed or something? It's supposed to run always when changing the CR, and at the point when the operand is about to get upgraded, the CR should already have the default values.
> 
> Nothing special. Just created one instance and upgraded it from one version to another. Nothing was disabled.

So after creating the first instance, before the upgrade, I assume the defaulter webhook was run normally.

> 
> > I did test the current code, and the defaulter webhook runs after the upgrade process - the upgrade patches the CR in-place, and at the time when it patches the CR in the cluster, the defaulter and validating webhook run automatically.
> 
> That was the problem we saw: it was run after the upgrade process. 

And it was also run after the upgrade.

> So, the reconciler was executed and, since the defaulter webhook was not run, one of the fields didn't have a value, making the operator crash.

When was the webhook not run, between upgrades?
The defaulter webhook doesn't run between upgrades, let's say an upgrade is from 0.1.0 (current) - 0.2.0 - 0.3.0 (latest), it won't run for 0.2.0. Was this the problem, do you have a link to the issue?

Afaics there is no way to run the 0.2.0-webhook-code because at the time of upgrade we only have the 0.3.0-webhook-code, but we could run the (latest) defaulter webhook code also for every upgrade if that makes sense. In the example above, run the webhook after the 0.2.0 upgrade.

andreasgerstmayr:
> That was the problem we saw: it was run after the upgrade process. So, the reconciler was executed and, since the defaulter webhook was not run, one of the fields didn't have a value, making the operator crash.

Reading it again...was the problem that the Reconciler was run before the upgrade? So the Reconciler tried to reconcile an object which had the default values of the previous version (as the old defaulter webhook was run), but not the default values of the current version (the current running defaulter webhook was not run)?

iblancasa:
> When was the webhook not run, between upgrades?
> The defaulter webhook doesn't run between upgrades, let's say an upgrade is from 0.1.0 (current) - 0.2.0 - 0.3.0 (latest), it won't run for 0.2.0. Was this the problem, do you have a link to the issue?

It was expected to run the 0.3.0 defaulter webhook for the instance to upgrade before the reconciliation.

This is the issue https://github.com/open-telemetry/opentelemetry-operator/issues/1799

> Reading it again...was the problem that the Reconciler was run before the upgrade? So the Reconciler tried to reconcile an object which had the default values of the previous version (as the old defaulter webhook was run), but not the default values of the current version (the current running defaulter webhook was not run)?

Was not.
1. The instance is created. Everything works
2. The operator is upgraded
3. The operator tries to reconcile the instance
4. One field is set to `nil`. It was expected as non-nil because the defaulter webhook sets it
5. Crash

I'm asking for this because talking to @pavolloffay we think is something we need to take into account. Maybe, if it is not possible to set something to run the defaulter webhook before the reconciliation for an upgraded instance, call the function manually or something (I don't really know how the implementation would be).

iblancasa:
> At the end of the upgrade process, the defaulter webhook (of 0.3.0, sticking with the version example in the comments above) runs, and after that the reconciler runs

In the OTEL Operator we didn't see that. We observed the upgrade process was run and later the reconciliation. That make the operator crash because some stuff was not set by the webhook. If for the Tempo Operator you tested the defaulter webhook runs before the reconciler, it's fine then.

andreasgerstmayr:
Yes, in my testing the upgrade process modifies the CR, which triggers the defaulter and validater webhook, and after that the reconciliation runs.

RELEASE.md
```diff
@@ -44,3 +44,19 @@ Upgrade to new version:
 ```
 IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z OPERATOR_NAMESPACE=openshift-operators make olm-upgrade
 ```
+
+## Test Upgrade using Manifests
+Install latest released version:
+```
+kubectl apply -f https://github.com/os-observability/tempo-operator/releases/latest/download/tempo-operator.yaml
```
pavolloffay, line 51: this points to the old organization

RELEASE.md
```diff
@@ -44,3 +44,19 @@ Upgrade to new version:
 ```
 IMG_PREFIX=docker.io/your_username OPERATOR_VERSION=x.y.z OPERATOR_NAMESPACE=openshift-operators make olm-upgrade
 ```
+
+## Test Upgrade using Manifests
```
pavolloffay, line 48: Is this mandatory part of the release process?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -220,12 +220,18 @@ type ComponentStatus struct {
 
 // TempoStackStatus defines the observed state of TempoStack.
 type TempoStackStatus struct {
+	// Version of the Tempo Operator.
+	// +optional
+	OperatorVersion string `json:"operatorVersion,omitempty"`
```
pavolloffay, line 225: Why do we need this field? 

Can there be multiple operator versions per instance in the same cluster?

apis/tempo/v1alpha1/tempostack_webhook.go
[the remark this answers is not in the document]
pavolloffay: Do you have any objections on keeping the label and respecting the managed CR option in the upgrade procedure? 

I would prefer to have a single way that mandates when reconcile/upgrade/validation is run.

pavolloffay (commented):
The OperatorVersion confuses me. 

Why the operator version is needed and it has to be exposed in the CR status? The operator should know which instances versions it can upgrade.

RELEASE.md
andreasgerstmayr: No, but I wanted to write it down somewhere.
Both manual upgrade steps (OLM and manifests) could be moved to a different file, CONTRIBUTING.md maybe?

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: This field is to identify which CRs need to be updated during the upgrade. It's required when we want to modify the CR without upgrading the tempo version.

The `tempoVersion` is only there for the end user, to have a nice output when running (`kubectl get tempo`).

andreasgerstmayr:
> The OperatorVersion confuses me.
> 
> Why the operator version is needed and it has to be exposed in the CR status? The operator should know which instances versions it can upgrade.

I think it makes sense to have versions for each software component - a version for the operator, and a version for tempo. While every tempo upgrade will involve an upgrade of the operator, not every new operator version will upgrade tempo. Let's say we modify the CRD but don't upgrade tempo - then we need a distinctive operator version.

Without this field, how does the operator know which upgrade steps it needs to execute? It only knows the current operator version (for example 0.5.0), but it cannot know at which operator version the CR was created (for example 0.3.0). With the operator version, the operator upgrade knows to run all upgrade steps between 0.3.0 and 0.5.0.

RELEASE.md
andreasgerstmayr: good point, updated.

RELEASE.md
andreasgerstmayr: I've moved it to CONTRIBUTING.md.

internal/upgrade/upgrade.go
```diff
@@ -0,0 +1,167 @@
...
+	return nil
+}
+
+// TempoStack upgrades a single TempoStack CR to the latest known version.
+// It runs all upgrade procedures between the current and latest operator version.
+// Note: It does not save/apply the changes to the CR.
+func (u Upgrade) TempoStack(ctx context.Context, tempo v1alpha1.TempoStack) (v1alpha1.TempoStack, error) {
+	log := u.Log.WithValues("namespace", tempo.Namespace, "tempo", tempo.Name)
+
+	if tempo.Spec.ManagementState == v1alpha1.ManagementStateUnmanaged {
```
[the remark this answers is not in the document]
pavolloffay: > Currently once they switch back and the operator is restarted

We should resolve this. We cannot wait for the operator to be restarted. Could you please book a ticket to resolve it and paste it here?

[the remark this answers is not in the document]
andreasgerstmayr: https://github.com/grafana/tempo-operator/issues/478

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -220,12 +220,14 @@ type ComponentStatus struct {
...
+
 	// Version of the managed Tempo instance.
 	// +optional
 	TempoVersion string `json:"tempoVersion,omitempty"`
-	// Version of the Tempo Query component used.
-	// +optional
-	TempoQueryVersion string `json:"tempoQueryVersion,omitempty"`
```
pavolloffay, line 228: This unfortunately seems like a breaking change https://github.com/grafana/tempo-operator/blob/v0.2.0/apis/tempo/v1alpha1/tempostack_types.go#L204

Renaming/removal of a field from CRD is a braking change. 

I would keep it here marked as deprecated and kept empty.

pavolloffay (commented):
LGTM apart removal of the queryversion from status

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: done

pavolloffay (approved):

PR #432, opened 2023-05-31
Rename operator deployment

andreasgerstmayr:
Commit 7fa448aa2bf28fb44ecf0cb530f7a530855894d6 added additional labels to the operator deployment and its selector field.

Unfortunately the selector field is immutable, and when OLM tries to patch the deployment while upgrading, Kubernetes will reject this update.

A workaround is to rename the deployment, as suggested here: https://github.com/operator-framework/operator-lifecycle-manager/issues/952

### DIFF
--- a/.chloggen/rename_operator_deployment.yaml
+++ b/.chloggen/rename_operator_deployment.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Rename operator deployment to enable upgrading from 0.1.0
+
+# One or more tracking issues related to the change
+issues: [432]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/rename_operator_deployment.yaml
+++ b/.chloggen/rename_operator_deployment.yaml
@@ -1,5 +1,5 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug_fix
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. operator, target allocator, github action)
 component: operator
@@ -13,4 +13,6 @@ issues: [432]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext:
+subtext: |
+  If you have installed the operator via Kubernetes manifests, please run `kubectl -n tempo-operator-system delete deployment tempo-operator-controller-manager` to prune the old deployment.
+  If you have installed the operator via OLM, no action is required.
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -8,7 +8,7 @@ metadata:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: controller-manager-v2
   namespace: system
   labels:
     control-plane: controller-manager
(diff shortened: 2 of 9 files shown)


config/manager/manager.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: controller-manager-v2
```
andreasgerstmayr, line 11: Does anyone have better suggestions? :)

iblancasa (commented):
I have some questions:
1. This is something we need to do just because the selector field was modified. Not something we'll need to do for each new version, right?
2. Are we going to bump the operator to version 2 and that is the reason for `v2` postfix?

Also... maybe it's ok to not support the upgrade from the `0.1.0` version. That will simplify everything

andreasgerstmayr:
> I have some questions:
> 
>     1. This is something we need to do just because the selector field was modified. Not something we'll need to do for each new version, right?

Yes, exactly. The selectors shouldn't change in the future.

>     2. Are we going to bump the operator to version 2 and that is the reason for `v2` postfix?

I choose v2 in a sense of "v2-of-the-deployment", not of the operator. The name of the operator deployment should be an internal detail and shouldn't matter much.

> Also... maybe it's ok to not support the upgrade from the `0.1.0` version. That will simplify everything

That's also an option. What are your concerns about renaming the deployment? Does anything rely on the name of the deployment (note, I didn't rename the labels of the deployment, only the name)?

Afaics nothing should break.. except one thing, if you install the operator via manifests, than the old deployment is not pruned. imho that's a general problem and will happen every time when we drop or rename any component (Service, ConfigMap, etc).

andreasgerstmayr:
> Also... maybe it's ok to not support the upgrade from the `0.1.0` version. That will simplify everything

Hm, yeah it also changes the pod name, then the blog post example output won't match anymore. Maybe it's not that important to support upgrades from 0.1.0, I don't know :)

iblancasa:
> That's also an option. What are your concerns about renaming the deployment? Does anything rely on the name of the deployment (note, I didn't rename the labels of the deployment, only the name)?
I think the `v2` can be a little bit misleading because we can think relates somehow to version 2 of the operator. Also, I'm not sure if we want to keep that `v2` postfix forever because of one change between the first and second versions.

It's not a strong opinion.

andreasgerstmayr:
I agree, I'm also not too happy about the rename with `-v2`. I'd like it renamed to just `tempo-operator` (I'm not a big fan of `controller-manager`), but that's also not possible because the kustomize files prefix everything with `tempo-operator`, so the deployment name in the kustomize files would be empty then.

.chloggen/rename_operator_deployment.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
frzifus, line 2:
```suggestion
change_type: breaking
```

.chloggen/rename_operator_deployment.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Rename operator deployment to enable upgrading from 0.1.0
+
+# One or more tracking issues related to the change
+issues: [432]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
frzifus, line 16: What actions are needed to overcome the breaking change? Maybe we should add an example here.

frzifus (commented):
Just some thoughts on it. ^^

config/manager/manager.yaml
andreasgerstmayr: Yep, that should work.

andreasgerstmayr: I've added instructions how to fix the breaking change to the changelog.

.chloggen/rename_operator_deployment.yaml
andreasgerstmayr: done

.chloggen/rename_operator_deployment.yaml
andreasgerstmayr: added

frzifus (approved):

PR #441, opened 2023-06-02
OLM upgrade test

andreasgerstmayr:
Resolves: #296


Makefile
```diff
@@ -340,6 +350,11 @@ prepare-e2e-openshift: deploy-minio
 e2e-openshift:
 	$(KUTTL) test --config kuttl-test-openshift.yaml
 
+# upgrade tests
+e2e-upgrade:
+	for f in tests/e2e-upgrade/*/*.envsubst; do envsubst < $$f > $${f%.envsubst}; done
```
pavolloffay, line 355: I would actually prefer to be explicit and call out files that need to be substituted with explicit variables.

pavolloffay (commented):
LGTM just minor comments

Makefile
```diff
@@ -278,6 +278,16 @@ endif
 catalog-build: opm ## Build a catalog image.
 	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
 
+.PHONY: fbc-catalog-build
+fbc-catalog-build: opm ## Build a File Based Catalog (FBC) catalog image
```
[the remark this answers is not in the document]
andreasgerstmayr: Currently nowhere.
I left it there because this will replace the catalog-build target soon-ish (see https://github.com/operator-framework/operator-sdk/issues/5832).

tests/e2e-upgrade/upgrade/.gitignore
```diff
@@ -0,0 +1,2 @@
+00-setup.yaml
```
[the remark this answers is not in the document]
andreasgerstmayr: Because this file is generated from `00-setup.yaml.envsubst`.

Unfortunately kuttl doesn't allow any parameterization of tests, therefore I made this workaround with envsubst.

Makefile
andreasgerstmayr: My goal with this was to use the same e2e test with different bundles, in particular downstream bundles.

For upstream (in the github action) we use
```
UPGRADE_FROM_CATALOG=quay.io/operatorhubio/catalog:latest
UPGRADE_TO_CATALOG=localregistry:5000/tempo-operator-catalog:v100.0.0
```
and for downstream we could invoke the test with something like `UPGRADE_TO_CATALOG=registry.internal.corp/iib:475490 make e2e-upgrade`

pavolloffay: That sounds good, my comment was more about changing the for loop and define which files needs substation, but it's a nit

pavolloffay: I would prefer to remove the parametrization for now, and add it once we have the use-case and start testing the upgrade on prow.ci cc) @IshwarKanse. 

We had issues with parametrization of kuttl tests in jaeger-operator, and I would like to avoid parametrization as much as possible.

IshwarKanse: To test operator upgrades using Prow CI, we can follow the steps outlined in the documentation (https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#operator-upgrade-testing), once all our e2e tests are running on Prow.

Currently, our OTEL and Tempo operator e2e tests run smoothly on Prow without any modifications. However, we are still working on finding a solution to run the Jaeger tests with the parameterization hacks. Therefore, I recommend avoiding the addition of parameterization until kuttl provides native support for it.

iblancasa (approved):

Makefile
andreasgerstmayr: This test needs the location of the new operator bundle - so running this test on Prow CI will fail in any case.
Unless Prow CI also runs the steps of the GitHub action to start a local registry with TLS and build and push the bundle and catalog there.

andreasgerstmayr: I've removed the parameterization for now.
Still interested how we can run the same test on prow ci as well (or later, as part of the QE process, with downstream bundles/catalogs).

IshwarKanse: There are steps to build the operator bundle and then upgrade the operator using the bundle image. However it will not use the github action like you said. I'll skip this testsuite from the Prow job and we can add a OpenShift specific one later.

Issue #363, opened 2023-03-30
olm-openshift: operands end up in crashloop


frzifus:
```bash
NAME                                             READY   STATUS             RESTARTS      AGE
minio-5794f6799d-pl6hn                           1/1     Running            0             25h
tempo-simplest-compactor-549ffd7b5c-z2db6        0/1     CrashLoopBackOff   3 (22s ago)   66s
tempo-simplest-distributor-5654468b76-llvcv      0/1     CrashLoopBackOff   3 (9s ago)    66s
tempo-simplest-gateway-67fd7f85bd-h7jhj          2/2     Running            0             65s
tempo-simplest-ingester-0                        0/1     CrashLoopBackOff   2 (32s ago)   66s
tempo-simplest-querier-fcbfbffd7-75pkl           0/1     CrashLoopBackOff   3 (9s ago)    66s
tempo-simplest-query-frontend-85bbb96575-wz7fj   0/2     CrashLoopBackOff   3 (21s ago)   66s
```

Used the following to create the tempo instance:
```yaml
apiVersion: tempo.grafana.com/v1alpha1
kind:  TempoStack
metadata:
  name: simplest
spec:
  storage:
    secret:
      name: minio-test
      type: s3
  storageSize: 1Gi
  resources:
    total:
      limits:
        memory: 2Gi
        cpu: 2000m
  tenants:
    mode: openshift
    authentication:
      - tenantName: dev
        tenantId: "1610b0c3-c509-4592-a256-a1871353dbfa"
      - tenantName: prod
        tenantId: "1610b0c3-c509-4592-a256-a1871353dbfb"
  template:
    gateway:
      enabled: true
    queryFrontend:
      jaegerQuery:
        enabled: true
```

All tempo logs show:
```bash
$ k logs -n summitdemo tempo-simplest-ingester-0
failed parsing config: failed to parse configFile /conf/tempo.yaml: yaml: unmarshal errors:
  line 52: field internal_server not found in type app.Config
```

Bundle was build using:
```bash
BUNDLE_VARIANT=openshift OPERATOR_VERSION=0.0.9 USER=frzifus make bundle docker-build docker-push docker-push bundle-build bundle-push catalog-build catalog-push
```

Version: https://github.com/os-observability/tempo-operator/commit/5ec7feaa09e5f4ce224732a148b1ca334c841d0c

---

Update:

**tempo.yaml**:
```bash
    48	      tls_min_version: VersionTLS13
    49	  search:
[...]

PR #450, opened 2023-06-07
Upgrade tempo to v2.1.1

andreasgerstmayr:
* deprecate maxSearchBytesPerTrace global and per-tenant limit
* add maxSearchDuration global and per-tenant limit

Resolves: #408
Resolves: #363


rubenvp8510 (approved):
LGTM

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -623,14 +623,23 @@ type QueryLimit struct {
...
 	// trace in bytes.
 	// default: `0` to disable.
 	//
 	// +optional
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Max Traces per User"
-	MaxSearchBytesPerTrace *int `json:"maxSearchBytesPerTrace"`
+	MaxSearchBytesPerTrace *int `json:"maxSearchBytesPerTrace,omitempty"`
+
+	// MaxSearchDuration defines the maximum search duration.
```
pavolloffay, line 636: Duration of what? The query execution in the Tempo engine or duration of a span/trace? 

Maybe something we should clarify?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -623,14 +623,23 @@ type QueryLimit struct {
...
+	// If this value is not set, then spec.search.maxDuration is used.
```
pavolloffay, line 637: What is the default value? Not set?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -623,14 +623,23 @@ type QueryLimit struct {
...
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Max Search Duration per User"
+	MaxSearchDuration metav1.Duration `json:"maxSearchDuration"`
```
pavolloffay, line 642: Why are we adding this field to the CR?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -623,14 +623,23 @@ type QueryLimit struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Max Tags per User"
 	MaxBytesPerTagValues *int `json:"maxBytesPerTagValues,omitempty"`
-	// MaxSearchBytesPerTrace defines the maximum size of search data for a single
+
+	// DEPRECATED. MaxSearchBytesPerTrace defines the maximum size of search data for a single
```
pavolloffay, line 627: Can we show a warning or error to the user when the option is used?

pavolloffay (commented):
LGTM, just some minor comments.

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: This field is per-tenant, `.spec.search.maxDuration` is a global config option.
If we don't want it configurable, I can remove it.

It's the best "replacement" for the deprecated `MaxSearchBytesPerTrace` option which I could find ;)

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: If the per-tenant `MaxSearchDuration` is unset, the global `spec.search.maxDuration` is used.

The default value for `spec.search.maxDuration` is noted in a comment of `maxDuration`: `default: 0s which means unlimited`.

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: It would be good, but where?

* operator logs during reconciling
* operator logs during an upgrade
* a new status condition
* restrict it in the validating webhook (that'd break the automated upgrade, but we can set it to 0 during the upgrade)

Any better ideas? I'm not sure what's a good place for it, except for the changelog/release notes.

apis/tempo/v1alpha1/tempostack_types.go
andreasgerstmayr: You're right, that comment wasn't very clear. I've updated it.

apis/tempo/v1alpha1/tempostack_types.go
pavolloffay: >operator logs during reconciling
> operator logs during an upgrade

I would avoid logs, usually users don't check them and even might not have permissions to do so.

I was thinking about validating webhook

andreasgerstmayr: Good idea, I've added validation to the webhook and added a upgrade procedure to remove this field if it is set.

pavolloffay (approved):

PR #453, opened 2023-06-07
Move route e2e test to OpenShift test suite

andreasgerstmayr:
Resolves: #350


Makefile
```diff
@@ -316,18 +316,14 @@ stop-kind:
 	$(ECHO)"Stopping the kind cluster"
 	$(VECHO)kind delete cluster
 
-.PHONY: install-openshift-routes
-install-openshift-routes:
-	./hack/install/install-openshift-routes.sh
```
iblancasa, line 321: Would it make sense to remove this file?

iblancasa (approved):

Makefile
andreasgerstmayr: I was thinking about it, but maybe it makes sense for local development/testing without crc

iblancasa: In that case, maybe, it would make sense to keep the makefile rule

frzifus (approved):

Issue #265, opened 2023-03-01
Existence and object type of gateway's ingress should be configurable


andreasgerstmayr:
- [ ] Ingress can be enabled/disabled per tempo (gateway) instance
- [ ] Add a option to use Ingress or Route

From https://github.com/os-observability/tempo-operator/pull/255#discussion_r1120257408

pavolloffay:
related to https://github.com/grafana/tempo-operator/issues/318

PR #479, opened 2023-06-25
Create ingress or Route for Gateway

rubenvp8510:
Resolves https://github.com/grafana/tempo-operator/issues/265


internal/manifests/gateway/openshift.go
```diff
@@ -86,23 +86,37 @@ func clusterRoleBinding(tempo v1alpha1.TempoStack) *rbacv1.ClusterRoleBinding {
...
+	case v1alpha1.TLSRouteTerminationTypeInsecure:
+		// NOTE: insecure, no tls cfg.
+	case v1alpha1.TLSRouteTerminationTypeEdge:
+		tlsCfg = &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge}
+	case v1alpha1.TLSRouteTerminationTypePassthrough:
+		tlsCfg = &routev1.TLSConfig{Termination: routev1.TLSTerminationPassthrough}
+	case v1alpha1.TLSRouteTerminationTypeReencrypt:
+		tlsCfg = &routev1.TLSConfig{Termination: routev1.TLSTerminationReencrypt}
+	default: // NOTE: if unsupported, end here.
+		return nil
```
iblancasa, line 101: Should we log something in this case? Because, for me, it sounds like failing silently

iblancasa (commented):
Should we add something to the E2E tests to verify this feature?

apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -146,6 +146,12 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 		r.Spec.Template.QueryFrontend.JaegerQuery.Ingress.Route.Termination = TLSRouteTerminationTypeEdge
 	}
 
+	// if tenant mode is Openshift, ingress type should be route by default.
+	if r.Spec.Tenants != nil && r.Spec.Tenants.Mode == OpenShift && r.Spec.Template.Gateway.Ingress.Type == IngressTypeRoute {
+		r.Spec.Template.Gateway.Ingress.Type = IngressTypeRoute
```
andreasgerstmayr, line 151: afaics `r.Spec.Template.Gateway.Ingress.Type == IngressTypeRoute` should be removed from the condition

internal/manifests/gateway/openshift.go
andreasgerstmayr: `TLSRouteTerminationType` is an enum in the CRD, so this case should never happen in theory.
Unless we update the list of valid enum values and forget to update this switch, so I guess yep, it makes sense to print a log message (could you add the same log message to the `switch` in the query_frontend also?).

apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -146,6 +146,12 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 		r.Spec.Template.QueryFrontend.JaegerQuery.Ingress.Route.Termination = TLSRouteTerminationTypeEdge
 	}
 
+	// if tenant mode is Openshift, ingress type should be route by default.
```
pavolloffay, line 149: @rubenvp8510 should there be a validation for jquery ingeress and gateway ingress? For instance do not allow Jquery ingress if the gateway is deployed?

pavolloffay (commented):
LGTM just a minor comment

internal/manifests/gateway/openshift.go
pavolloffay: It should be checked by link https://github.com/grafana/tempo-operator/blob/7c7ee6470cd01138f1225d8734dd7f3e7229135f/.golangci.yaml#L29

iblancasa (approved):

apis/tempo/v1alpha1/tempostack_webhook.go
rubenvp8510: Yes, that is true, we need to validate it.

rubenvp8510: Indeed I think that validation already happens here: https://github.com/grafana/tempo-operator/blob/main/apis/tempo/v1alpha1/tempostack_webhook.go#L254

pavolloffay (approved):

PR #487, opened 2023-06-28
Add ability to enable multitenancy without gateway

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/224

### DIFF
--- a/internal/manifests/gateway/gateway.go
+++ b/internal/manifests/gateway/gateway.go
@@ -37,10 +37,6 @@ const (
 
 // BuildGateway creates gateway objects.
 func BuildGateway(params manifestutils.Params) ([]client.Object, error) {
-	if !params.Tempo.Spec.Template.Gateway.Enabled {
-		return []client.Object{}, nil
-	}
-
 	rbacCfg, tenantsCfg, err := buildConfigFiles(newOptions(params.Tempo, params.Gates.OpenShift.BaseDomain, params.GatewayTenantSecret, params.GatewayTenantsData))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


iblancasa (commented):
Some unit tests are failing.

Also: do we want to add some kind of step to the E2E tests to verify this or do we think it will be enough with the unit tests?

internal/manifests/gateway/gateway.go
```diff
@@ -37,8 +37,7 @@ const (
 
 // BuildGateway creates gateway objects.
 func BuildGateway(params manifestutils.Params) ([]client.Object, error) {
-	if !params.Tempo.Spec.Template.Gateway.Enabled ||
-		params.Tempo.Spec.Tenants == nil {
+	if !params.Tempo.Spec.Template.Gateway.Enabled {
```
andreasgerstmayr, line 40: Personally I wouldn't call `BuildGateway()` in `BuildAll()` if `params.Tempo.Spec.Template.Gateway.Enabled == false`

andreasgerstmayr (commented):
LGTM once the unit tests pass

rubenvp8510:
> Some unit tests are failing.
> 
> Also: do we want to add some kind of step to the E2E tests to verify this or do we think it will be enough with the unit tests?

Not sure, let me think if we can add an e2e test.

andreasgerstmayr (approved):

rubenvp8510:
Wait a little bit until merge this, I'd like to do a final test :)

rubenvp8510:
This is ready to merge, I've already test it, not sure how to add

Issue #488, opened 2023-06-28
Use reconcile.TerminalError for errors requiring human intervention (e.g. missing storage secret)


andreasgerstmayr:
See https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0
Requires #448

PR #500, opened 2023-07-03
Refactor status handling

andreasgerstmayr:
* rename Degraded status condition to ConfigurationError to cleary indicate the type of error
* return `reconcile.TerminalError` for terminal errors requiring human intervention (e.g. ConfigurationErrors)
* add a new status reason `FailedReconciliation` for generic errors, e.g. permission errors during reconciliation, and show the error message
* collect all errors during applying and pruning objects
* remove `requeue` return value from `status.Refresh()` because it was always true if error was set, and therefore redundant
* drop unused `StatusClient` from `status.*Condition()`

Resolves: #488

### DIFF
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -247,22 +247,22 @@ type TempoStackStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
-// ConditionStatus defines the status of a condition (e.g. ready or degraded).
+// ConditionStatus defines the status of a condition (e.g. ready, failed, pending or configuration error).
 type ConditionStatus string
 
 const (
 	// ConditionReady defines that all components are ready.
 	ConditionReady ConditionStatus = "Ready"
-	// ConditionDegraded defines that one or more components are in a degraded state.
-	ConditionDegraded ConditionStatus = "Degraded"
 	// ConditionFailed defines that one or more components are in a failed state.
 	ConditionFailed ConditionStatus = "Failed"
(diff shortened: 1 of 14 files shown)


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -247,22 +247,22 @@ type TempoStackStatus struct {
...
 
-// ConditionStatus defines the status of a condition (e.g. ready or degraded).
+// ConditionStatus defines the status of a condition (e.g. ready, failed, pending or configuration error).
 type ConditionStatus string
 
 const (
 	// ConditionReady defines that all components are ready.
 	ConditionReady ConditionStatus = "Ready"
-	// ConditionDegraded defines that one or more components are in a degraded state.
-	ConditionDegraded ConditionStatus = "Degraded"
```
pavolloffay, line 257: This seems like a breaking change, we should probably deprecate the degraded state.

andreasgerstmayr: In the CRD we declare `Conditions []metav1.Condition` in `TempoStackStatus`, so from a strongly-typed-standpoint I'd argue it's not a breaking change. The operator can put "random" `metav1.Condition`s in this slice.

With the current code, in case the TempoStack is never in a degraded state, the Degraded condition won't be in this slice of status conditions. It gets only appended if the Degraded state is set currently (and once it was set, it stays in the slice).

So, if I re-add this definition, effectively nothing<sup>1</sup> will change, as the operator will never get into the degraded state anymore.
<sup>1</sup> The only thing changes is the metrics, we'll have a `tempostack_status_condition{condition="Degraded"}` metric which is always `0`.

pavolloffay (approved):

PR #501, opened 2023-07-04
enable TLS and builtInCertManagement

rubenvp8510:
This depends on https://github.com/grafana/tempo-operator/pull/450


rubenvp8510:
@andreasgerstmayr Do you think we should enable mTLS for both openshift and community? I think in that way we can have the feature more tested.

andreasgerstmayr:
> @andreasgerstmayr Do you think we should enable mTLS for both openshift and community? I think in that way we can have the feature more tested.

I think it's good to have security by default, so I'm +1 on also enabling it in the community version.

.chloggen/enable_tls.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable mTLS for internal components
```
andreasgerstmayr, line 8:
```suggestion
note: Enable mTLS by default for internal components
```

I'd add "by default" so it's clear this change is just a change to the default settings, and that it can also be disabled.

andreasgerstmayr (approved):

andreasgerstmayr:
@rubenvp8510 did you run the e2e tests on OpenShift? Most likely they'll also need some updates in the `*-assert.yaml` files.

andreasgerstmayr (changes requested):
The OpenShift e2e suite needs to be updated for the new volumes containing the TLS certs.

rubenvp8510:
> @rubenvp8510 did you run the e2e tests on OpenShift? Most likely they'll also need some updates in the `*-assert.yaml` files.

Ohh yes, I forgot that one. thanks!

rubenvp8510:
Openshift tests updated!

andreasgerstmayr (approved):

Issue #490, opened 2023-06-28
verify-traces kuttl step should check number of traces using Jaeger API


andreasgerstmayr:
Instead of just checking if any trace exists.

pavolloffay:
Preferably use tempo APIs and move jaeger/3rd party testing to a separate suite.

Issue #98, opened 2022-10-24
Reduce test times for e2e tests


kevinearls:
This is a follow up to #91   In that test I ran tracegen for 30 seconds to avoid a race condition in the verification step where traces may not yet be available.

If it's not possible to fix this via configuration then maybe add a retry loop to the verification step.

pavolloffay:
The 30seconds: https://github.com/grafana/tempo-operator/blob/edec6e294884dd2d38a81d392e2c3112d81478fa/tests/e2e/smoketest-with-jaeger/02-generate-traces.yaml#L15

pavolloffay:
+1 we can consider a shorter time or report a specific amount of data

PR #509, opened 2023-07-12
Verify number of traces in verify-traces steps

andreasgerstmayr:
* speed up telemetrygen by generating 10 traces instead of running for 30 seconds
* use Tempo API where applicable
* install minio storage before each test which writes or reads from storage (to get deterministic results)

Resolves: #98
Resolves: #490


iblancasa (approved):
Maybe, it would make sense to move the "verification of traces" logic to one script and mount it in a volume or something. Or to build our own image to run the tests. Just as a way to make it easier to read the tests files and avoid copying/pasting if we need to modify something.

Also... would it make sense to pin the `registry.gitlab.com/gitlab-ci-utils/curl-jq` image to an specific version? I suppose there should not be problems and should be something easy to fix just in case.

pavolloffay (approved):
@andreasgerstmayr did you run the tests on OCP?

andreasgerstmayr:
> @andreasgerstmayr did you run the tests on OCP?

Yes

PR #514, opened 2023-07-17
Scope PrometheusRule to a specific TempoStack instance

### LINKED ISSUE #485: Scope PrometheusRule to a specific TempoStack instance
When enabling `.spec.observability.metrics.createPrometheusRules` multiple alerts are generated, for example:
```
      - alert: TempoDistributorUnhealthy
        annotations:
          message: There are {{ printf "%f" $value }} unhealthy distributor(s).
          runbook_url: https://github.com/grafana/tempo/tree/main/operations/tempo-mixin/runbook.md#TempoDistributorUnhealthy
        expr: |
          max by (cluster, namespace) (tempo_ring_members{state="Unhealthy", name="distributor", namespace=~".*"}) > 0
        for: 15m
        labels:
          severity: warning
```

These alerts match Tempo instance of the entire cluster, but they should only match Tempo instances of the TempoStack where `createPrometheusRules` was enabled.

Suggested solution: add `cluster="$tempostack_name", namespace="$namespace"` to every alert.
Note: Also update the recording rules.

rubenvp8510:
Sign-off-by: Ruben Vargas <ruben.vp8510@gmail.com>

Fixes #485


internal/manifests/alerts/prometheus-alerts.yaml
```diff
@@ -111,7 +111,7 @@ groups:
       message: "Ingesters in {{ $labels.cluster }}/{{ $labels.namespace }} are receiving more data/second than desired, add more ingesters."
       runbook_url: "[[ .RunbookURL ]]#TempoProvisioningTooManyWrites"
     expr: |
-      avg by (cluster, namespace) (rate(tempo_ingester_bytes_received_total{job=~".+/ingester"}[1m])) / 1024 / 1024 > 30
+      avg by (cluster, namespace) (rate(tempo_ingester_bytes_received_total{cluster="[[ .Cluster ]]", namespace="[[ .Namespace ]]", ob=~".+/ingester"}[1m])) / 1024 / 1024 > 30
```
andreasgerstmayr, line 114:
```suggestion
      avg by (cluster, namespace) (rate(tempo_ingester_bytes_received_total{cluster="[[ .Cluster ]]", namespace="[[ .Namespace ]]", job=~".+/ingester"}[1m])) / 1024 / 1024 > 30
```

internal/manifests/alerts/prometheus-rules.yaml
```diff
@@ -2,15 +2,15 @@
...
-  - expr: "sum(rate(tempo_request_duration_seconds_sum[1m])) by (cluster, namespace, job, route) / sum(rate(tempo_request_duration_seconds_count[1m])) by (cluster, namespace, job, route)"
-    record: "cluster_namespace_job_route:tempo_request_duration_seconds:avg"
-  - expr: "sum(rate(tempo_request_duration_seconds_bucket[1m])) by (le, cluster, namespace, job, route)"
-    record: "cluster_namespace_job_route:tempo_request_duration_seconds_bucket:sum_rate"
-  - expr: "sum(rate(tempo_request_duration_seconds_sum[1m])) by (cluster, namespace, job, route)"
-    record: "cluster_namespace_job_route:tempo_request_duration_seconds_sum:sum_rate"
-  - expr: "sum(rate(tempo_request_duration_seconds_count[1m])) by (cluster, namespace, job, route)"
-    record: "cluster_namespace_job_route:tempo_request_duration_seconds_count:sum_rate"
+  - expr: "histogram_quantile(0.99, sum(rate(tempo_request_duration_seconds_bucket{cluster=\"[[ .Cluster ]]\", namespace=\"[[ .Namespace ]]\"}[1m])) by (le, cluster, namespace, job, route))"
+    record: "[[ .Cluster ]]_[[ .Namespace ]]:cluster_namespace_job_route:tempo_request_duration_seconds:99quantile"
```
andreasgerstmayr, line 6: Is it required to update the name (`record`) of the new metrics?

In my cluster I have the old rules (before this PR) and I see the following metrics:
```
cluster_namespace_job_route:tempo_request_duration_seconds:99quantile{cluster="simplest", job="default/compactor", namespace="default", route="metrics"}
```
So the recorded metrics already contain the `namespace` and `cluster` labels, so I *think* (I didn't test it) it'd be fine to only change the `expr` (as you did) but leave the `record` the same, wdyt?

rubenvp8510: If I don't scope the `records`, those are overridden

If I create a "simplest" `Tempostack`, this is how the record looks like: 
Screenshot from 2023-07-17 22-22-41

Then after create simplest2, this is the only records appears on Prometheus.

Screenshot from 2023-07-17 22-20-42

andreasgerstmayr: I cannot reproduce this behavior - I just tried it on my cluster (using Prometheus 2.44.0), when I open the `/rules` page I see two sections `tempo_alerts` and two sections for `tempo_rules` (one with `simplest`, one with `simplest2`).

When I query for a recorded metric, for example `cluster_namespace_job_route:tempo_request_duration_seconds:99quantile`, I see values for both TempoStack instances: Bildschirmfoto vom 2023-07-18 13-00-49

Can you check this page again to be sure you don't have two `tempo_rules` sections? (fwiw, maybe we could also rename `tempo_rules` and `tempo_alerts` to `tempo_rules_$cluster_$namespace` and `tempo_alerts_$cluster_$namespace`?)

rubenvp8510: I was able to reproduce what you have, I was confused by the fact the groups are the same, but I have a question about what happens when you refer a record in the alerts section and two records has the same name?

andreasgerstmayr (approved):

PR #521, opened 2023-07-24
Publish API docs only as a md file

pavolloffay:
The API docs will be published only as .md files and referenced from https://grafana.com/docs/tempo/next/setup/operator/

### DIFF
--- a/docs/operator/api.md
+++ b/docs/operator/api.md
@@ -13,7 +13,6 @@ This Document contains the types introduced by the Tempo Operator to be consumed
 
 <b>Resource Types:</b>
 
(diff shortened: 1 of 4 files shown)


iblancasa (approved):

frzifus (approved):

docs/operator/api.md
```diff
@@ -4227,6 +4170,5 @@ string
 
 <hr/>
 
-
-
++newline
```
andreasgerstmayr, line 4173: Is `+newline` some kind of marker and supposed to be here?

PR #532, opened 2023-07-31
Fix defaulter for openshift, allow create ingress

rubenvp8510:
Note that this get rid of the possibility of disable ingress on the gateway on openshift, because if empty string for the type, we set it to route, only if the openshift feature gate is enabled.


apis/tempo/v1alpha1/tempostack_webhook.go
```diff
@@ -137,17 +137,19 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
...
-
 	// if tenant mode is Openshift, ingress type should be route by default.
-	if r.Spec.Tenants != nil && r.Spec.Tenants.Mode == OpenShift {
+	if r.Spec.Tenants != nil && r.Spec.Tenants.Mode == OpenShift && r.Spec.Template.Gateway.Ingress.Type == "" {
 		r.Spec.Template.Gateway.Ingress.Type = IngressTypeRoute
-		r.Spec.Template.Gateway.Ingress.Route.Termination = TLSRouteTerminationTypePassthrough
 	}
 
+	if r.Spec.Template.Gateway.Ingress.Type == IngressTypeRoute && r.Spec.Template.Gateway.Ingress.Route.Termination == "" {
+		r.Spec.Template.Gateway.Ingress.Route.Termination = TLSRouteTerminationTypeEdge
```
pavolloffay, line 146: Why are we defaulting to edge termination?

pavolloffay: nit: It might be useful to define a constant `defaultRouteTLSTermination`

rubenvp8510: Done with the constant recommendation.

Actually not sure about an specific reason to use edge termination, I just take that default from the other ingress, also I think was the default value used when there was no configuration on the gateway ingress.

andreasgerstmayr (approved):

apis/tempo/v1alpha1/tempostack_webhook.go
andreasgerstmayr: When I test the changes of this PR on OpenShift, I'm getting `Client sent an HTTP request to an HTTPS server.` from the gateway, because the gateway expects a HTTPS connection. With `edge`, the connection is terminated at the router and the connection from the router to the pod is unencrypted.

The OpenShift e2e tests also fail now, because we used to have `passthrough`: https://github.com/grafana/tempo-operator/blob/2e973db9e56cc9f12928d9c19bf50fe4634ac3c8/tests/e2e-openshift/multitenancy/01-assert.yaml#L266-L267

andreasgerstmayr (changes requested):
`edge` termination breaks the connection to the gateway: `Client sent an HTTP request to an HTTPS server.`
We need `passthrough` TLS termination for the gateway.

rubenvp8510:
> `edge` termination breaks the connection to the gateway: `Client sent an HTTP request to an HTTPS server.` We need `passthrough` TLS termination for the gateway.

Ohh you're right, now I see my error

andreasgerstmayr (approved):

PR #557, opened 2023-08-15
fix: provide correct mode name via operator-sdk annotation for oc console

### LINKED ISSUE #556: Unsupported mode value when creating tempostack form openshift console
**Description of the problem**

When creating a TempoStack from the OpenShift webconsole via the form view, only static and dynamic options are available in Tenants Configuration - Mode. If we select the dynamic option, we get the following error.  Also the Tenants Configuration Mode is a required field. 

```
Error "Unsupported value: "dynamic": supported values: "static", "openshift"" for field "spec.tenants.mode". 
```
**Version of components**

tempo-operator.v0.1.0-6

**OCP Server Version** 
4.13.0-0.nightly-2023-05-25-001936

**How reproducible**

Always

**Steps to reproduce the issue**

- Install the Tempo Operator.

- Go to OCP Webconsole > Installed Operators > Tempo Operator > TempoStack. Click on Create TempoStack and check in the Tenants Configuration in the Form View only static and dynamic options are available. If we try to create the TempoStack we get the below error. 

image

frzifus:
closes #556 

Any idea how to create a good test case for it?

### DIFF
--- a/apis/tempo/v1alpha1/auth.go
+++ b/apis/tempo/v1alpha1/auth.go
@@ -22,7 +22,7 @@ type TenantsSpec struct {
 	// +required
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default:=static
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:static","urn:alm:descriptor:com.tectonic.ui:select:dynamic"},displayName="Mode"
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:static","urn:alm:descriptor:com.tectonic.ui:select:openshift"},displayName="Mode"
 	Mode ModeType `json:"mode"`
 
 	// Authentication defines the tempo-gateway component authentication configuration spec per tenant.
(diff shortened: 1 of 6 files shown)


andreasgerstmayr (approved):

apis/tempo/v1alpha1/auth.go
```diff
@@ -22,7 +22,7 @@ type TenantsSpec struct {
 	// +required
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default:=static
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:static","urn:alm:descriptor:com.tectonic.ui:select:dynamic"},displayName="Mode"
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:static","urn:alm:descriptor:com.tectonic.ui:select:openshift"},displayName="Mode"
```
andreasgerstmayr, line 25: do we actually need this, shouldn't openshift pick up the enum definition here:
```
// +kubebuilder:validation:Enum=static;openshift
type ModeType string
```
?

frzifus: I would say yes, to compare it on other places in the code.

andreasgerstmayr: I just removed the `+operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors....` line and can't see any difference in OpenShift:
image

PR #569, opened 2023-08-24
[WIP] Set env vars to allow disable TLS per component

rubenvp8510:
Before fix the unit  tests I would like to know what do you think on this solution

@andreasgerstmayr @iblancasa @pavolloffay @frzifus

### DIFF
--- a/internal/manifests/queryfrontend/query_frontend.go
+++ b/internal/manifests/queryfrontend/query_frontend.go
@@ -118,6 +118,7 @@ func deployment(params manifestutils.Params) (*v1.Deployment, error) {
 								"-target=query-frontend",
 								"-config.file=/conf/tempo.yaml",
 								"-mem-ballast-size-mbs=1024",
+								"-config.expand-env=true",
 							},
 							Ports: []corev1.ContainerPort{
 								{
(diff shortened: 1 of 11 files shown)


iblancasa (commented):
For me is fine

internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -145,6 +146,24 @@ func deployment(params manifestutils.Params) (*v1.Deployment, error) {
 							},
 							Resources:       manifestutils.Resources(tempo, manifestutils.QueryFrontendComponentName),
 							SecurityContext: manifestutils.TempoContainerSecurityContext(),
+							Env: []corev1.EnvVar{
+								{
+									Name:  "HTTP_CERT_FILE",
+									Value: " ",
```
andreasgerstmayr, line 152: Does Tempo trim this whitespace in the env var and interpret it as a missing value?

rubenvp8510: I had to put this whitespace because if not tempo will interpret the variable as not defined and set the default value, but this whitespace has the same effect as put nothing on the tls config, and disabled it.

May be what I can do is to get rid of default value and use env vars in all components, to make this thing uniform WDYT?

andreasgerstmayr: > May be what I can do is to get rid of default value and use env vars in all components, to make this thing uniform WDYT?

Yeah, I think that would make it less hacky :)

I just remembered that if we want to have separate configs, we don't have to create two ConfigMaps, because a ConfigMap can have multiple keys. So alternatively we could just call `buildConfiguration(params)` [twice](https://github.com/grafana/tempo-operator/blob/2f1ffba9e35d1fe4f2b34381f8fc1886bfdd8600/internal/manifests/config/configmap.go#L22) with some new parameter, add `tempo-query-frontend.yaml` [here](https://github.com/grafana/tempo-operator/blob/2f1ffba9e35d1fe4f2b34381f8fc1886bfdd8600/internal/manifests/config/configmap.go#L40) and calculate the checksum over all files (tempo, tempo-query and tempo-query-frontend; not overrides.yaml, because tempo reloads that file automatically).

rubenvp8510: Oh yes, you're right, may be is better to call buildConfiguration twice, for me sounds less hacky than the env vars solution. I'll modify this then.

Issue #561, opened 2023-08-16
Enable mTLS again if Grafana works with and without gateway


andreasgerstmayr:
Revert #560, but make sure Grafana works with and without gateway (see #552)

PR #576, opened 2023-08-25
Disable mTLS for query frontend when gateway is disabled

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/561


iblancasa (commented):
Just asking because I don't know if it would make sense: do we want to test as part of the E2E tests what happens when the encryption feature flags are disabled? Or are we ok with just the unit tests?

pavolloffay (approved):

rubenvp8510:
> Just asking because I don't know if it would make sense: do we want to test as part of the E2E tests what happens when the encryption feature flags are disabled? Or are we ok with just the unit tests?

I think is a good idea, but we need to figure out how to do it in the right way. I would say this could be a separate PR.

iblancasa (approved):

rubenvp8510:
@andreasgerstmayr Fixed the openshift tests

```
--- PASS: kuttl (154.20s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/route (19.82s)
        --- PASS: kuttl/harness/multitenancy (125.48s)
        --- PASS: kuttl/harness/monitoring (144.18s)
        --- PASS: kuttl/harness/red-metrics (147.59s)

```

andreasgerstmayr (approved):

rubenvp8510:
Any other thing required to merge this?

Issue #527, opened 2023-07-27
Support configuration of TLS in receiver settings 


frzifus:
Support configuring TLS on the distributor.
See: https://grafana.com/docs/tempo/latest/configuration/tls/#receiver-tls

I would like to enable TLS on the OpenTelemetry end to ship traces into tempo. 

Example exporter config:
```yaml
exporters:
  # Export the application tenant traces to a Tempo instance
  otlp/application:
    endpoint: tempo-observatorium-tempostack-distributor:4317
    tls:
      insecure: false
      ca_file: "/etc/pki/ca-trust/source/service-ca/service-ca.crt"
```

cc @andreasgerstmayr

pavolloffay:
+1

pavolloffay:
We should as well think about supporting TLS on the gateway for data ingestion

PR #577, opened 2023-08-26
Support configuration of TLS in receiver settings

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/527


internal/manifests/config/tempo-config.yaml
```diff
@@ -30,6 +51,13 @@ distributor:
             key_file: {{ .TLS.Paths.Key }}
             min_version: {{ .TLS.Profile.MinTLSVersionShort }}
 {{- end }}
+{{- if .ReceiverTLS.Enabled }}
```
andreasgerstmayr, line 54: If `Gates.GRPCEncryption .Gateway` and `.ReceiverTLS.Enabled` is enabled, we end up with two `tls:` sections. Can you add a validation to the webhook to prevent enabling `ReceiverTLS` if `Gates.GRPCEncryption .Gateway` is enabled?

internal/manifests/config/tempo-config.yaml
```diff
@@ -30,6 +51,13 @@ distributor:
...
+          tls:
+            ca_file:  {{ .ReceiverTLS.Paths.CA }}
+            cert_file: {{ .ReceiverTLS.Paths.Certificate }}
+            key_file: {{ .ReceiverTLS.Paths.Key }}
+            min_version: {{ .ReceiverTLS.MinTLSVersion }}
+{{- end }}
 {{- if not .Gateway }}
         http:
           endpoint: 0.0.0.0:4318
```
andreasgerstmayr, line 63: OTLP/HTTP also supports TLS according to https://grafana.com/docs/tempo/latest/configuration/tls/

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -116,6 +116,8 @@ type TempoStackSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Observability"
 	Observability ObservabilitySpec `json:"observability,omitempty"`
+
+	ReceiversTLS ReceiversTLSSpec `json:"receiversTLS,omitempty"`
```
andreasgerstmayr, line 120: I suggest to put this configuration under `.spec.template.distributor.tls` (TLS config of the distributor component) instead of a new top-level `.spec.receiversTLS`.

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -387,6 +389,48 @@ type ObjectStorageTLSSpec struct {
...
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret",displayName="Certificate Secret Name"
+	Cert string `json:"certName,omitempty"`
+	// certName is the name of a Secret containing a certificate and the private key
+	// It needs to be in the same namespace as the Tempo custom resource.
+	//
+	// +optional
+	// +kubebuilder:validation:optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Key of the secret for private key"
+	CertPrivateKey string `json:"certPrivateKey,omitempty"`
```
andreasgerstmayr, line 417: Personally I'd hardcode `tls.key` and `tls.crt`, afaics it's a Kubernetes convention to use these key names: https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets
But I don't have a very strong opinion on that.

andreasgerstmayr: (and also hardcode `ca.crt`, but I couldn't find a convention for that)

andreasgerstmayr:
LGTM, I've added a few comments

apis/tempo/v1alpha1/tempostack_types.go
rubenvp8510: Make sense to me, I haven't found also something for the ca, but I see other parts of this code use ca.crt so I think we will be fine.

frzifus (approved):

PR #581, opened 2023-08-29
Automate release process

rubenvp8510:
I do believe that bumping the tempo version should be part of another process, may be an issue or ticket
, not part of the release. We can have a release for fixing minor operator issues that doesn't require to bump tempo.

This PR automates the release, create changelog, create a commit on main with the generated changelog and the bundles. 

In order to trigger this process we can go go Actions -> Tab -> Prepare Release action and trigger it manually. It receives the version of tempo operator we want to release.

See this screenshot:

Screenshot from 2023-08-30 14-59-37

This was tested on my fork.

@pavolloffay is this what you are looking for?

Will address the OLM publish in a separate PR.


pavolloffay:
@rubenvp8510 we need to document this in the release docs

rubenvp8510:
> @rubenvp8510 we need to document this in the release docs

Done

andreasgerstmayr (approved):
Awesome :rocket: 

One question: Is this workflow idempotent?
In case something fails during the workflow (some timeout, GH action issue, any random issue), can we re-run it?

andreasgerstmayr:
Could you also revert the hardcoded operator version in the Makefile: https://github.com/grafana/tempo-operator/blob/84c0ea1535c7ed03b63889a92196c521b72254d1/Makefile#L4, it used to be
```
OPERATOR_VERSION ?= $(or $(shell git describe --tags --abbrev=0 2> /dev/null | sed 's/^v//'), 0.0.0)
```

I think we reverted back to the hardcoded version because the changelog PR was failing, but we don't have that PR anymore now, so afaics it should work.

rubenvp8510:
> Awesome 🚀
> 
> One question: Is this workflow idempotent? In case something fails during the workflow (some timeout, GH action issue, any random issue), can we re-run it?

Excelent question, I think everything here is idempotent, except the final commit. May be we should send the PR and not commit direct to main? In this way we can see the result before mergin?

andreasgerstmayr:
> May be we should send the PR and not commit direct to main? In this way we can see the result before mergin?

Yeah, I was thinking the same, then we could also modify the changelog if we want to. However, then we have to manually tag the commit, as we can't have a PR which auto-tags itself afaics (maybe another follow-up workflow which gets triggered after merging the changelog-PR? :laughing:).

rubenvp8510:
> > May be we should send the PR and not commit direct to main? In this way we can see the result before mergin?
> 
> Yeah, I was thinking the same, then we could also modify the changelog if we want to. However, then we have to manually tag the commit, as we can't have a PR which auto-tags itself afaics (maybe another follow-up workflow which gets triggered after merging the changelog-PR? 😆).

I think is in the way it works now, when you merge the PR, the tag and the GH release are created.

.github/workflows/release.yaml
```diff
@@ -11,9 +15,25 @@ jobs:
...
+        OPERATOR_VERSION: ${{inputs.version}}
+        IMG_PREFIX: ghcr.io/grafana/tempo-operator
+
+    - name: Generate CHANGELOG
+      run: make chlog-update
+      env:
+        OPERATOR_VERSION: ${{inputs.version}}
+
+    # Commit all changed files back to the repository
+    - uses: stefanzweifel/git-auto-commit-action@v4
```
andreasgerstmayr, line 32: On a second thought, could we replace this with plain `git` commands? In terms of securing our build pipeline, I think it's best to avoid any third-party actions/components. Maybe at some point this external repo gets compromised and then we would ship malicious code in all future releases.

afaics this action just does the following (untested):
```
git commit -m "Release ${{inputs.version}}"
git push
git tag v${{inputs.version}}
git push v${{inputs.version}}
```

rubenvp8510: Changed

.idea/tempo-operator.iml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andreasgerstmayr, line 1: I think the `.idea` folder should be added to `.gitignore`

.github/workflows/release.yaml
andreasgerstmayr: Perfect, thanks! :100:

andreasgerstmayr (approved):

PR #592, opened 2023-09-06
Use tempo SA to query RED metrics from OpenShift monitoring stack

### LINKED ISSUE #526: Use tempo service account for accessing user-workload monitoring stack
The #470 adds support for the monitor tab and on OpenShift it uses thanos-query as the metrics store. 

On OpenShift the #470 requires that a user creates `ClusterRoleBinding` to give  user permission to query metrics from Thanos e.g. The user token is sent to Thanos (instead of tempo-query SA) because ` --query.bearer-token-propagation=true` is set on tempo-query.

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: jaeger-cluster-monitoring-view
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-monitoring-view
subjects:
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: system:authenticated
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: system:unauthenticated  
```

The proposal is to send tempo SA as token to Thanos and let the operator to create the binding:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: tempo-cluster-monitoring-view
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-monitoring-view
subjects:
  - kind: ServiceAccount
    name: tempo-simplest
    namespace: ploffay
```

This will as well require change on Jaeger codebase to make sure Tempo token is set even when `--query.bearer-token-propagation=true` is set - see https://github.com/jaegertracing/jaeger/blob/9c787fca2dfcbe4182de85d34f36ba6cd18bbb7e/plugin/metrics/prometheus/metricsstore/reader.go#L323

pavolloffay:
Resolves #526 

Depends on https://github.com/jaegertracing/jaeger/pull/4726 and released tempo-query with the feature.

Notable changes:
- tempo operator SA on OCP gets `cluster-monitoring-view` cluster role - it needs it because it needs to create the same `clusterrolebinding` for instances.
- tempo instance SA gets `cluster-monitoring-view` cluster role - needed to access metrics for the monitor tab

### DIFF
--- a/bundle/openshift/manifests/tempo-operator-manager-rolebinding-cluster-monitoring-view_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
+++ b/bundle/openshift/manifests/tempo-operator-manager-rolebinding-cluster-monitoring-view_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
@@ -1,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  creationTimestamp: null
+  labels:
+    app.kubernetes.io/managed-by: operator-lifecycle-manager
+    app.kubernetes.io/name: tempo-operator
+    app.kubernetes.io/part-of: tempo-operator
+  name: tempo-operator-manager-rolebinding-cluster-monitoring-view
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-monitoring-view
+subjects:
+- kind: ServiceAccount
+  name: tempo-operator-controller-manager
+  namespace: openshift-operators-redhat
--- a/config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
+++ b/config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: manager-rolebinding-cluster-monitoring-view
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-monitoring-view
+subjects:
+  - kind: ServiceAccount
+    name: controller-manager
+    namespace: system
--- a/config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
+++ b/config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
@@ -1,3 +1,5 @@
+# The operator needs cluster-monitoring-view role
+# because it creates the same binding for the operands
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/internal/manifests/queryfrontend/query_frontend.go
+++ b/internal/manifests/queryfrontend/query_frontend.go
@@ -83,7 +83,8 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 		}
 	}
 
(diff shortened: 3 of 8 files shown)


bundle/openshift/manifests/tempo-operator-manager-rolebinding-cluster-monitoring-view_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  creationTimestamp: null
```
andreasgerstmayr, line 4: let's remove this line

bundle/openshift/manifests/tempo-operator-manager-rolebinding-cluster-monitoring-view_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,17 @@
...
+    app.kubernetes.io/part-of: tempo-operator
+  name: tempo-operator-manager-rolebinding-cluster-monitoring-view
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-monitoring-view
+subjects:
+- kind: ServiceAccount
+  name: tempo-operator-controller-manager
+  namespace: openshift-operators-redhat
```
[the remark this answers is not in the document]
andreasgerstmayr: Ah sorry, that's the generated file.

config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
andreasgerstmayr, line 1: Can you add a comment to explain why the operator needs this privilege (so it can grant it to the individual Tempo CR Service Accounts)?

internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -80,25 +83,31 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 		}
 	}
 
+	if tempo.Spec.Template.QueryFrontend.JaegerQuery.Enabled && tempo.Spec.Template.QueryFrontend.JaegerQuery.MonitorTab.Enabled &&
+		tempo.Spec.Template.QueryFrontend.JaegerQuery.MonitorTab.PrometheusEndpoint == thanosQuerierOpenShiftMonitoring {
```
andreasgerstmayr, line 87: To make it easier for customers, should we set this value by default on OpenShift in the defaulter webhook, if the MonitorTab is enabled?

simonpasquier:
lgtm

config/overlays/openshift/role_binding_cluster_monitoring_view.yaml
pavolloffay: done

internal/manifests/queryfrontend/query_frontend.go
pavolloffay: Let's do it separately https://github.com/grafana/tempo-operator/issues/618

andreasgerstmayr (approved):

PR #609, opened 2023-09-14
[chore] Build and publish an image to replace curl-jq for the E2E tests

### LINKED ISSUE #608: Replace the curl-jq by one that can be build for multiple architectures
The image `registry.gitlab.com/gitlab-ci-utils/curl-jq` is built for amd64. Having one image we can build for other architectures will be really useful for our internal testing process.

iblancasa:
Related to #608

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -325,7 +325,12 @@ deploy-minio:
 
 # generic end-to-tests
 .PHONY: prepare-e2e
-prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy
+prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy test-tools-image
+
+.PHONY: test-tools-image
+test-tools-image:
+	docker build -t local/ci-tools:e2e -f tests/Dockerfile.utils tests
+	kind load docker-image local/ci-tools:e2e
 
 .PHONY: e2e
 e2e:
--- a/Makefile
+++ b/Makefile
@@ -325,12 +325,7 @@ deploy-minio:
 
 # generic end-to-tests
 .PHONY: prepare-e2e
-prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy test-tools-image
-
-.PHONY: test-tools-image
-test-tools-image:
-	docker build -t local/ci-tools:e2e -f tests/Dockerfile.utils tests
-	kind load docker-image local/ci-tools:e2e
+prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy
 
 .PHONY: e2e
 e2e:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -325,7 +325,12 @@ deploy-minio:
 
 # generic end-to-tests
 .PHONY: prepare-e2e
-prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy
+prepare-e2e: kuttl start-kind cert-manager deploy-minio set-test-image-vars build docker-build load-image-operator deploy test-tools-image
```
andreasgerstmayr, line 328: If we can, I would avoid this additional dependency of building a container image and pushing it to the correct registry (in the Makefile `kind` is used, but what about prow.ci, etc.?) for running the e2e tests.

This container has no dependencies on the tempo-operator - I'd suggest we publish this image with all required architectures in the github container registry (maybe a new github action which pushes the image weekly), and reference it from there in the tests? Then it'll work in all test setups without any additional work.

Or am I missing something, and there's some advantage of having it always built before running the e2e tests?

iblancasa: > If we can, I would avoid this additional dependency of building a container image and pushing it to the correct registry (in the Makefile `kind` is used, but what about prow.ci, etc.?) for running the e2e tests.

Other systems don't use the {prepare-e2e} make target because it starts kind, installs the cert manager, and other stuff. Those systems run their own automation and I added this rule as any other step.

> This container has no dependencies on the tempo-operator - I'd suggest we publish this image with all required architectures in the github container registry (maybe a new github action which pushes the image weekly), and reference it from there in the tests? Then it'll work in all test setups without any additional work.

Why weekly?

> Or am I missing something, and there's some advantage of having it always built before running the e2e tests?

I think this image should not change too much but... if we need to add a package, for instance, we will need to send one PR for the change in the image and another one for the real work. But yes, I can add the automation to publish it.

iblancasa: I modified the PR to build and publish the image. As soon as the image is built and publish, I'll create a second PR to use it in the E2E tests.

andreasgerstmayr: > > If we can, I would avoid this additional dependency of building a container image and pushing it to the correct registry (in the Makefile `kind` is used, but what about prow.ci, etc.?) for running the e2e tests.
> 
> Other systems don't use the {prepare-e2e} make target because it starts kind, installs the cert manager, and other stuff. Those systems run their own automation and I added this rule as any other step.

Yep, my idea was that we can avoid requiring any changes to these other systems, which don't use the `prepare-e2e` Makefile step.

> 
> > This container has no dependencies on the tempo-operator - I'd suggest we publish this image with all required architectures in the github container registry (maybe a new github action which pushes the image weekly), and reference it from there in the tests? Then it'll work in all test setups without any additional work.
> 
> Why weekly?

Just to keep the image up-to-date with latest CVE fixes etc. But in this case it probably doesn't matter much I guess, as we control the way we invoke the container anyway.

> 
> > Or am I missing something, and there's some advantage of having it always built before running the e2e tests?
> 
> I think this image should not change too much but... if we need to add a package, for instance, we will need to send one PR for the change in the image and another one for the real work. But yes, I can add the automation to publish it.

I think it's worth it, this image will hardly change, but we'll run our e2e tests countless times in multiple systems (locally, prow.ci, CVP pipeline, etc) and an extra step to get the tests running is imho not required.

andreasgerstmayr (approved):

PR #627, opened 2023-09-21
Exposing the Tempo API using the gateway

rubenvp8510:
This depends on https://github.com/observatorium/api/pull/575


internal/manifests/gateway/gateway.go
```diff
@@ -168,6 +168,8 @@ func deployment(params manifestutils.Params, rbacCfgHash string, tenantsCfgHash
 								fmt.Sprintf("--traces.write.endpoint=%s:4317", naming.ServiceFqdn(tempo.Namespace, tempo.Name, manifestutils.DistributorComponentName)),
 								fmt.Sprintf("--traces.read.endpoint=%s://%s:16686", httpScheme(params.Gates.HTTPEncryption),
 									naming.ServiceFqdn(tempo.Namespace, tempo.Name, manifestutils.QueryFrontendComponentName)),
+								fmt.Sprintf("--traces.tempo.endpoint=%s://%s:3200", httpScheme(params.Gates.HTTPEncryption),
```
andreasgerstmayr, line 171: Let's use `manifestutils.PortHTTPServer` for the Tempo port, and also the relevant port constant from `manifestutils/constants.go` for the 2 other hardcoded ports above.

pavolloffay:
Don't forget to bump the gateway version :)

rubenvp8510:
@andreasgerstmayr Do you think we need to have a flat feature or a field in the CRD to enable or disable this feature?

andreasgerstmayr:
> @andreasgerstmayr Do you think we need to have a flag feature or a field in the CRD to enable or disable this feature?

I think we don't need a new flag, because we already expose the same data (all traces) with a different API (tempo-query).

andreasgerstmayr (commented):
Did you run the openshift e2e tests? Afaics the multitenancy test should fail, as it has an assert on the container arguments.
Can you also update gateway_test.go to add an assert for the new traces.tempo.endpoint CLI argument?

LGTM otherwise :+1:

rubenvp8510:
> Did you run the openshift e2e tests? Afaics the multitenancy test should fail, as it has an assert on the container arguments. Can you also update gateway_test.go to add an assert for the new traces.tempo.endpoint CLI argument?
> 
> LGTM otherwise 👍

Oh yes you're right, I'll update openshift tests! Thanks for the review

rubenvp8510:
@andreasgerstmayr fixed openshift test, also added verify traces for the traceql endpoint. and the assert of the argument

andreasgerstmayr (approved):
awesome, thanks! I'll merge after CI passes.

rubenvp8510:
Thank youu!

Issue #591, opened 2023-09-05
Set RELATED_IMAGE env variables in operator deployment


andreasgerstmayr:
Required for disconnected environments to detect images the operator depends on.

Docs:
https://docs.openshift.com/container-platform/4.13/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs
https://redhat-connect.gitbook.io/certified-operator-guide/troubleshooting-and-resources/offline-enabled-operators#golang-operators

andreasgerstmayr:
Also drop `images` from configuration file and `image` field from TempoStack CR

PR #638, opened 2023-10-06
Set RELATED_IMAGE environment variables in operator deployment

andreasgerstmayr:
* add RELATED_IMAGE env vars to operator deployment
* remove images field from operator configuration

Resolves: #591


bundle/community/manifests/tempo-operator.clusterserviceversion.yaml
```diff
@@ -887,6 +896,15 @@ spec:
   maturity: alpha
   provider:
     name: Grafana Tempo Operator SIG
+  relatedImages:
```
andreasgerstmayr, line 899: :raised_hands:

config/manager/manager.yaml
```diff
@@ -38,6 +38,15 @@ spec:
         image: controller:latest
         args:
         - --leader-elect
+        env:
```
pavolloffay, line 41: It would be good to document how versions should be updated

andreasgerstmayr: I've added a section to `CONTRIBUTING.md`.

rubenvp8510:
This looks good to me, I think we need to fix the conflicts first.

andreasgerstmayr:
I've resolved the conflicts now

Issue #409, opened 2023-05-15
Fix rendering of tempo configmap


andreasgerstmayr:
The `tempo.yaml` in the configmap gets rendered as a string:
```
$ kubectl get configmap tempo-simplest -o yaml                                                        
apiVersion: v1
data:
  overrides.yaml: |
    overrides:
  tempo-query.yaml: |
    backend: 127.0.0.1:3200
    tenant_header_key: x-scope-orgid
  tempo.yaml: "compactor:\n  compaction:\n    block_retention: 48h0m0s\n  ring:\n
    \   kvstore:\n      store: memberlist\ndistributor:\n  receivers:\n    jaeger:\n
    \     protocols:\n        thrift_http:\n          endpoint: 0.0.0.0:14268\n    otlp:\n
    \     protocols:\n        grpc:\n          endpoint: 0.0.0.0:4317\n        http:\n
    \         endpoint: 0.0.0.0:4318\n  ring:\n    kvstore:\n      store: memberlist\ningester:\n
    \ lifecycler:\n    ring:\n      kvstore:\n        store: memberlist\n      replication_factor:
    1\n    tokens_file_path: /var/tempo/tokens.json\n  max_block_duration: 10m\nmemberlist:\n
    \ abort_if_cluster_join_fails: false\n  join_members:\n  - tempo-simplest-gossip-ring\nmultitenancy_enabled:
    false\noverrides:\n  max_search_bytes_per_trace: 0\nquerier:\n  max_concurrent_queries:
    20\n  frontend_worker:\n    frontend_address: tempo-simplest-query-frontend-discovery:9095\n
    \ search:\n    external_hedge_requests_at: 8s\n    external_hedge_requests_up_to:
    2\nserver:\n  grpc_server_max_recv_msg_size: 4194304\n  grpc_server_max_send_msg_size:
    4194304\n  http_listen_port: 3200\n  http_server_read_timeout: 3m\n  http_server_write_timeout:
    3m\n  log_format: logfmt\n  log_level: debug\nstorage:\n  trace:\n    backend:
    s3\n    blocklist_poll: 5m\n    cache: none\n    azure:\n      container_name:
    \n    gcs:\n      bucket_name: \n    s3:\n      endpoint: minio.minio.svc:9000\n
    \     bucket: tempo\n      insecure: true\n    local:\n      path: /var/tempo/traces\n
[...]

PR #639, opened 2023-10-06
Check for trailing whitespace in configmap

andreasgerstmayr:
If there is a trailing whitespace in the configmap, the configmap will be rendered in a suboptimal way (see #409).


andreasgerstmayr:
Tests pass now, no trailing whitespace anymore :)

rubenvp8510 (approved):

Issue #674, opened 2023-11-09
Automated rebuilds won't pick up new images


andreasgerstmayr:
Various tools are capable of automatic rebuilding of images in case of CVEs (e.g. [freshmaker](https://github.com/redhat-exd-rebuilds/freshmaker)). These tools rebuild all images and update the `RELATED_IMAGE_` environment variables in the CSV accordingly (ref. #638).

However, they **do not** update the version field (the `github.com/grafana/tempo-operator/internal/version.operatorVersion` LDFLAG when building the operator), they only update the version of the CSV. Therefore the upgrade process won't run, and nothing will replace the `images` field in the TempoStack CRs, i.e. the tempo/tempo-query/gateway pods are not updated with the new image version.

imho the best solution is to keep the `images` field unset if it is the default. In this case, the operator will use the image from `RELATED_IMAGE_` env variables, and the containers are always using the latest images as specified in the `RELATED_IMAGE_` env vars. As a bonus, if a user is manually editing the images field, it won't get reset on every upgrade.

PR #675, opened 2023-11-09
Unset (default) images in TempoStack CR

andreasgerstmayr:
Various tools are capable of automatic rebuilding of images in case of
CVEs (e.g. freshmaker [1]). These tools rebuild all images and update
the RELATED_IMAGE_ environment variables in the CSV accordingly (ref. #638).

However, they do not update the version field (the
github.com/grafana/tempo-operator/internal/version.operatorVersion LDFLAG
when building the operator), they only update the version of the CSV.
Therefore the upgrade process won't run, and nothing will replace the
images field in the TempoStack CRs, i.e. the tempo/tempo-query/gateway
pods are not updated with the new image version.

This PR modifies the handling to use the default images if the image
field of the TempoStack CR is unset. Then the operator will use the image
from RELATED_IMAGE_ env variables, and the containers are always using the
latest images as specified in the RELATED_IMAGE_ env vars. As a bonus,
if a user is manually editing the images field, it won't get reset on every
upgrade.

[1] https://github.com/redhat-exd-rebuilds/freshmaker

Resolves: #674

### DIFF
--- a/.chloggen/unset_images_in_cr.yaml
+++ b/.chloggen/unset_images_in_cr.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Unset (default) images in TempoStack CR
+
+# One or more tracking issues related to the change
+issues: [674]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: From 0.6.0 onwards, the image location is not stored in the TempoStack CR unless it got changed manually
--- a/.chloggen/unset_images_in_cr.yaml
+++ b/.chloggen/unset_images_in_cr.yaml
@@ -13,4 +13,6 @@ issues: [674]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: From 0.6.0 onwards, the image location is not stored in the TempoStack CR unless it got changed manually
+subtext: |
+  This upgrade reverts any change to the `spec.images` fields of any TempoStack instance.
+  Beginning with version 0.6.0, the image location is not stored in the TempoStack instance unless it is changed manually.
--- a/internal/upgrade/v0_6_0.go
+++ b/internal/upgrade/v0_6_0.go
@@ -1,0 +1,17 @@
+package upgrade
+
+import (
+	"context"
+
+	"github.com/grafana/tempo-operator/apis/tempo/v1alpha1"
+)
+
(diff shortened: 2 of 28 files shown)


internal/upgrade/v0_6_0.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"context"
+
+	"github.com/grafana/tempo-operator/apis/tempo/v1alpha1"
+)
+
+// This upgrade unsets the image fields in the TempoStack CR.
+// From 0.6.0 onwards, the image location is not stored in the CR unless it got changed manually.
+func upgrade0_6_0(ctx context.Context, u Upgrade, tempo *v1alpha1.TempoStack) (*v1alpha1.TempoStack, error) {
+	tempo.Spec.Images.Tempo = ""
```
rubenvp8510, line 12: If I understand this correctly, this is going to unset all the images on the CRs, but if an user specified an image and then do the upgrade, he will lose that setting?

andreasgerstmayr: Thinking more about it. If the user configures a custom image, let's say a rebuild with modifications of tempo 2.2.3, and then upgrades the tempo operator and the new operator version requires tempo version 2.4.
This will likely break the installation anyway?

andreasgerstmayr: In other words, setting a custom image only makes sense if the `TempoStack` is set to unmanaged imho (unmanaged also means it won't get upgraded).

rubenvp8510: I don't think it is a way, I think the approach here is the right one. May be we need to put in the changelog or in some place that if you upgrade from 0.5.0 to 0.6.0 and you have custom images those changes will be overridden.

rubenvp8510: Yes, that make sense to me.

andreasgerstmayr: +1, I updated the text in the changelog entry to make it more clear.

rubenvp8510 (approved):

andreasgerstmayr:
fwiw, opentelemetry-operator (https://github.com/open-telemetry/opentelemetry-operator/blob/d76cf668d95bf70fff8e473cfb163f35d500b911/internal/manifests/collector/container.go#L40-L43) and jaeger-operator (https://github.com/jaegertracing/jaeger-operator/blob/50cd8c138b6f0b7de6463097069aa6f024b7e0bb/pkg/util/util.go#L277-L287) follow the same pattern

frzifus (approved):
Thats actually quite nice :)

.chloggen/unset_images_in_cr.yaml
```diff
@@ -0,0 +1,18 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Unset (default) images in TempoStack CR
+
+# One or more tracking issues related to the change
+issues: [674]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
pavolloffay, line 16: I would document here the workaround

pavolloffay (approved):

.chloggen/unset_images_in_cr.yaml
andreasgerstmayr: Which workaround?

PR #708, opened 2023-12-12
[Chore] Add cluster role binding required for OpenShift monitoring dependent cases.

IshwarKanse:
The Thanos Querier has switched to using kube-rbac-proxy in OpenShift 4.15 which requires creation of additional role and binding to query for metrics. https://issues.redhat.com/browse/MON-3379 This PR adds the required ClusterRole and ClusterRolebinding so that the in-cluster monitoring dependent cases work on OCP 4.15.


tests/e2e-openshift/monitoring/04-verify-traces.yaml
```diff
@@ -1,3 +1,28 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: kuttl-cluster-monitoring-metrics-api
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: kuttl-cluster-monitoring-metrics-api
+subjects:
+- kind: ServiceAccount
+  name: prometheus-user-workload
```
andreasgerstmayr, line 22: The tempo operator does not own/manage the `prometheus-user-workload` service account, is it really the responsibility of the tempo operator (tests) to grant this service account additional permissions?

Shouldn't this permission be granted in the CMO operator?

andreasgerstmayr: > The ClusterRoleBinding step is not a Tempo requirement and is used only by the test case to check the metrics using the check_metrics.sh script.

Ah! Thanks for the explanation. That was the part I didn't realize, that check_metrics.sh is impersonating the `prometheus-user-workload` service account.
Could you add this as a comment above the ClusterRoleBinding?

IshwarKanse: Added the comment. :)

andreasgerstmayr: Thank you!

andreasgerstmayr (approved):

PR #711, opened 2023-12-13
Generate a commented, full example CR

andreasgerstmayr:
Generate an commented, full example CR with default values in the `api-docs` Makefile target.
This helps navigating through the nested objects of the CR without having to jump back and forth between objects (either in the source code or the generated [api.md](https://github.com/grafana/tempo-operator/blob/main/docs/operator/api.md)).


pavolloffay (approved):

pavolloffay:
Can we get rid of the old API docs?

pavolloffay:
Could you please submit docs here https://grafana.com/docs/tempo/latest/setup/operator/api-docs/

andreasgerstmayr:
The current generator only supports CRDs (generated from `tempo.grafana.com_tempostacks.yaml`), so it doesn't support creating docs for the operator config (the `ProjectConfig` struct).

I'll create a follow-up issue to migrate the operator config and remove the old docs.

PR #712, opened 2023-12-13
Configure the number of replicas for compactor, querier and query-frontend according to the CR

andreasgerstmayr:
Configure the number of replicas for compactor, querier and query-frontend according to the CR

### DIFF
--- a/apis/tempo/v1alpha1/tempostack_types.go
+++ b/apis/tempo/v1alpha1/tempostack_types.go
@@ -103,7 +103,6 @@ type TempoStackSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tempo Component Templates"
 	Template TempoTemplateSpec `json:"template,omitempty"`
 
-	// NOTE: currently this field is not considered.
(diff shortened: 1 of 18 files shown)


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -103,7 +103,6 @@ type TempoStackSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tempo Component Templates"
 	Template TempoTemplateSpec `json:"template,omitempty"`
 
-	// NOTE: currently this field is not considered.
```
andreasgerstmayr, line 106: cc @rubenvp8510 this note is outdated, right? The setting of this field lands in the tempo config:
```
    ingester:
      lifecycler:
        ring:
          replication_factor: 2
```

rubenvp8510: Yes it is outdated, I might forgot to remove it when this was implemented.

rubenvp8510 (approved):

PR #722, opened 2023-12-21
Support monolithic deployment mode

### LINKED ISSUE #710: Support Tempo deployments in monolithic mode
The operator should support Tempo deployments in [monolithic mode](https://grafana.com/docs/tempo/latest/setup/deployment/#monolithic-mode).
For example local filesystem and in-memory storage is only available in monolithic deployments.

Design doc: https://docs.google.com/document/d/1fbBfXddfukj6Pfkr8rtRfUj522pIk5JKBzA_WyppPW0

```[tasklist]
### Tasks
- [x] In-memory storage: #722
- [x] PV storage: #722
- [x] Object storage: #763
- [x] Ingestion OTLP/gRPC: #722
- [x] Ingestion OTLP/HTTP: #722
- [x] Ingestion TLS: #756
- [x] Jaeger UI: #722
- [x] Jaeger UI Ingress+Route: #755
- [x] Multitenancy/Gateway: #816
- [x] Observability: ServiceMonitor, PrometheusRules, GrafanaDatasource: #793
- [x] Overlay config: #722
- [x] Management State: #722
- [x] Resources/Limits: #771
- [x] Operand status field: #787
- [x] Upgrade: #850
- [x] Tolerations (#782)
- [x] support custom service account: #836
- [x] TLS Profile: #862
- [ ] watch external resources (storage secret, T config map and secret)
```

andreasgerstmayr:
Support Tempo monolithic deployment mode with a new `TempoMonolithic` CR.
Partially resolves #710.

- [X] In-memory storage
- [X] PV storage
- [X] Ingestion OTLP/gRPC
- [X] Ingestion OTLP/HTTP
- [X] Jaeger UI
- [X] Overlay config
- [X] Management State
- [x] Tests


.env
pavolloffay, line 1: Why is this file needed? When are the tags updated?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,265 @@
...
+// MonolithicObservabilitySpec defines the observability settings of the Tempo deployment.
+type MonolithicObservabilitySpec struct {
+	// Metrics defines the metrics configuration of the Tempo deployment
+	//
+	// +kubebuilder:validation:Optional
+	Metrics *MonolithicObservabilityMetricsSpec `json:"metrics,omitempty"`
+}
+
+// MonolithicObservabilityMetricsSpec defines the metrics settings of the Tempo deployment.
+type MonolithicObservabilityMetricsSpec struct {
```
pavolloffay, line 201: Can't we reuse smth from the microservice type? 

e.g. the whole observability spec https://github.com/grafana/tempo-operator/blob/cbe369b4983893cf50ce60201a1ed055f71bc344/apis/tempo/v1alpha1/tempostack_types.go#L141

apis/tempo/v1alpha1/tempomonolithic_webhook.go
```diff
@@ -0,0 +1,77 @@
...
+// SetupWebhookWithManager will setup the manager to manage the webhooks.
+func (r *TempoMonolithic) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type.
+func (r *TempoMonolithic) Default() {
+	log := ctrl.Log.WithName("tempomonolithic-webhook")
```
pavolloffay, line 19: how is this rendered in the logs? Isn't it too long?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// MonolithicTracesStorageBackend defines the backend storage for traces.
+//
+// +kubebuilder:validation:Enum=memory;pv
+type MonolithicTracesStorageBackend string
+
+const (
+	// MonolithicTracesStorageBackendMemory specifies a in-memory storage backend.
+	MonolithicTracesStorageBackendMemory MonolithicTracesStorageBackend = "memory"
```
pavolloffay, line 75: I think we should explain more what does the in-memory mean as there is no true in-memory storage in tempo.

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,265 @@
...
+	// MonolithicTracesStorageBackendPersistentVolume specifies a Persistent Volume storage backend.
+	MonolithicTracesStorageBackendPersistentVolume MonolithicTracesStorageBackend = "pv"
```
pavolloffay, line 77: will S3 be supported?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,265 @@
...
+	// +kubebuilder:validation:Required
+	Traces MonolithicTracesStorageSpec `json:"traces"`
+}
+
+// MonolithicTracesStorageSpec defines the traces storage for the Tempo deployment.
+type MonolithicTracesStorageSpec struct {
+	// Backend defines the backend for storing traces
+	//
+	// +kubebuilder:validation:Required
+	Backend MonolithicTracesStorageBackend `json:"backend"`
```
pavolloffay, line 55: What is the default?

docs/operator/api.md
pavolloffay, line 1: why there are so many changes in this file?

docs/spec/tempomonolithic.yaml
pavolloffay, line 1: How is this file generated? Is the tool for generating part of the repo tooling and CI check for changes? 

I would prefer to introduce it in a separate PR and align the microservices one with this as well

andreasgerstmayr: I've updated #711 to include the generation in the `api-docs` Makefile target

.env
andreasgerstmayr: It's currently only used for the `run` Makefile target. I'm also not happy with the duplication of `.env` and `config/manager/manager.yaml`. Maybe we should remove the values from `manager.yaml` and use the ones from `.env` for it?

Or add these default image locations to the top of the Makefile?

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Yes, in a follow-up PR.

docs/operator/api.md
andreasgerstmayr: afaics just a bad diff algorithm, it detects insertions as edits.
This file contains a alphabetic-sorted list of all used structs in our CRDs, and the monolithic spec adds a couple more structs, inserting them between the other already existing ones. And the diff algorithm thinks it was an edit rather than an insert.

apis/tempo/v1alpha1/tempomonolithic_webhook.go
andreasgerstmayr: It'll print this line:
```
{"level":"debug","ts":"2024-01-15T19:07:08.746058013+01:00","logger":"tempomonolithic-webhook","msg":"running defaulter webhook","name":"sample"}
```
if debug logs are enabled (`go run ./main.go --zap-log-level=debug start`).

andreasgerstmayr: But as we'll set the defaults in the reconcile loop now, I'm removing this log statement.

I'll keep it there for the validating webhook, which is still in use.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: I've updated the comment

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: `memory`, I've updated the comment and marker.

.env
andreasgerstmayr: I've extracted this change to #746

.chloggen/monolithic_mode.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support monolithic deployment mode
+
+# One or more tracking issues related to the change
+issues: [710]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay, line 16: Can we have more details here? This is important and cool new feature. It would be good to mention the new type and memory storage.

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,266 @@
...
+type MonolithicStorageSpec struct {
+	// Traces defines the backend storage configuration for traces
+	//
+	// +kubebuilder:validation:Required
+	Traces MonolithicTracesStorageSpec `json:"traces"`
+}
+
+// MonolithicTracesStorageSpec defines the traces storage for the Tempo deployment.
+type MonolithicTracesStorageSpec struct {
+	// Backend defines the backend for storing traces. Default: memory
```
pavolloffay, line 52: I am a bit concerned about using the in-memory as default. 

The upstream uses PV as default and the in-memory can be easily overlooked and use significant resources in the cluster.

controllers/tempo/tempomonolithic_controller.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		// we'll ignore not-found errors, since they can't be fixed by an immediate
+		// requeue (we'll need to wait for a new notification), and we can get them
+		// on deleted requests.
+		return ctrl.Result{}, nil
+	}
+
+	// apply defaults
+	tempo.Default()
```
pavolloffay, line 59: Is the default webhook registered as well?

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: If you're only concerned about OOM situations, the memory counts towards the container resource limit:
> While tmpfs is very fast be aware that, unlike disks, files you write count against the memory limit of the container that wrote them

https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

I'd prefer to keep memory as default, as it's great for quick testing/demos/showcases (and is the default for jaeger all-in-one), and changing it to `pv` is easy (and can/should be mentioned in the docs).

controllers/tempo/tempomonolithic_controller.go
andreasgerstmayr: No, only the validating webhook is registered.
I'm already running with `make run` locally, much faster feedback cycle :)

.chloggen/monolithic_mode.yaml
andreasgerstmayr: Good idea, I've updated the entry.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: That would be inconsistent with the
```
<feature>:
  enabled: true
```
style of the CR. I'd prefer to migrate the `TempoStack`, maybe in the next CRD version? Shouldn't be too difficult to create a conversion webhook for this.

```
spec:                                    # TempoMonolithicSpec defines the desired state of TempoMonolithic.
  observability:                         # Observability defines observability configuration for the Tempo deployment
    metrics:                             # Metrics defines the metrics configuration of the Tempo deployment
      prometheusRules:                   # ServiceMonitors defines the PrometheusRule configuration
        enabled: false                   # Enabled defines if the operator should create PrometheusRules for this Tempo deployment
      serviceMonitors:                   # ServiceMonitors defines the ServiceMonitor configuration
        enabled: false                   # Enabled defines if the operator should create ServiceMonitors for this Tempo deployment
```

pavolloffay: so we agreed on using the enabled field which is better supported in tools like `kustomize` and `kubectl edit` (the empty structs are removed). Are we going to reuse some parts from the monolithic APIs?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// ManagementState defines whether this instance is managed by the operator or self-managed
+	//
+	// +kubebuilder:validation:Optional
+	Management ManagementStateType `json:"management,omitempty"`
+
+	// Observability defines observability configuration for the Tempo deployment
+	//
+	// +kubebuilder:validation:Optional
+	Observability *MonolithicObservabilitySpec `json:"observability,omitempty"`
```
pavolloffay, line 34: Asking again, can we reuse some of the structs from the microservies type?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,269 @@
...
+	// +kubebuilder:validation:Required
+	Enabled bool `json:"enabled"`
+
+	// CA defines the name of a secret containing the CA certificate
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	CA string `json:"ca"`
+
+	// Cert defines the name of a secret containing the TLS certificate and private key
```
pavolloffay, line 155: what are the entries in the secret?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,269 @@
...
+type MonolithicJaegerUISpec struct {
+	// Enabled defines if the Jaeger UI should be enabled
+	//
+	// +kubebuilder:validation:Required
+	Enabled bool `json:"enabled"`
+
+	// Ingress defines the ingress configuration for Jaeger UI
+	//
+	// +kubebuilder:validation:Optional
+	Ingress *MonolithicJaegerUIIngressSpec `json:"ingress,omitempty"`
```
pavolloffay, line 172: Can be the ingress spec reused from the microservices? There are more settings users might want to configure

controllers/tempo/tempomonolithic_controller.go
pavolloffay: We should document it on the defaulting webhook

internal/manifests/mutate.go
```diff
@@ -12,10 +13,22 @@ import (
...
 
+// ImmutableErr occurs if an immutable field should be changed.
+type ImmutableErr struct {
+	field    string
+	existing interface{}
+	desired  interface{}
+}
+
+func (m *ImmutableErr) Error() string {
+	return fmt.Sprintf("update to immutable field %s is forbidden", m.field)
```
pavolloffay, line 29: shouldn't it print the existing and desired as well?

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Sorry, I thought I removed this struct already. I'll use a generic TLS struct from TempoStack in the next PR where I add TLS for ingestion.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Yes, we can reuse a TLS struct, the multitenancy structs, ManagementStateType, LimitSpec and the storage secret.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: We can, question is do we value _consistency inside the same CR_ or _consistency between the two CRs_ more?

```
spec:
  observability:                         # Observability defines observability configuration for the Tempo deployment
    metrics:                             # Metrics defines the metrics configuration of the Tempo deployment
      prometheusRules:                   # ServiceMonitors defines the PrometheusRule configuration
        enabled: false                   # Enabled defines if the operator should create PrometheusRules for this Tempo deployment
      serviceMonitors:                   # ServiceMonitors defines the ServiceMonitor configuration
        enabled: false
```

vs

```
spec:
  observability:                         # ObservabilitySpec defines how telemetry data gets handled.
    grafana:                             # Grafana defines the Grafana configuration for operands.
      createDatasource: false            # CreateDatasource specifies if a Grafana Datasource should be created for Tempo.
      instanceSelector:                  # InstanceSelector specifies the Grafana instance where the datasource should be created.
    metrics:                             # Metrics defines the metrics configuration for operands.
      createPrometheusRules: false       # CreatePrometheusRules specifies if Prometheus rules for alerts should be created for Tempo components.
      createServiceMonitors: false       # CreateServiceMonitors specifies if ServiceMonitors should be created for Tempo components.
    tracing:                             # Tracing defines a config for operands.
      jaeger_agent_endpoint: "localhost:6831" # JaegerAgentEndpoint defines the jaeger endpoint data gets send to.
      sampling_fraction: ""
```

andreasgerstmayr: The first example is consistent with the rest of the Monolithic CR, and allows additional settings for prometheusRules or serviceMonitors in the future.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Already ready in #755:
```
spec:
  jaegerui:                              # JaegerUI defines the Jaeger UI configuration
    enabled: false                       # Enabled defines if the Jaeger UI should be enabled
    ingress:                             # Ingress defines the ingress configuration for Jaeger UI
      annotations:                       # Annotations defines the annotations of the Ingress object.
        "key": ""
      enabled: false                     # Enabled defines if an Ingress object should be created for Jaeger UI
      host: ""                           # Host defines the hostname of the Ingress object.
      ingressClassName: ""               # IngressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource.
    route:                               # Route defines the route configuration for Jaeger UI
      annotations:                       # Annotations defines the annotations of the Ingress object.
        "key": ""
      enabled: false                     # Enabled defines if a Route object should be created for Jaeger UI
      host: ""                           # Host defines the hostname of the Ingress object.
      ingressClassName: ""               # IngressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource.
      termination: "edge"                # Termination specifies the termination type. Default: edge.
```

vs spec of TempoStack:
```
      jaegerQuery:                       # JaegerQuerySpec defines Jaeger Query specific options.
        enabled: false                   # Enabled is used to define if Jaeger Query component should be created.
        ingress:                         # Ingress defines Jaeger Query Ingress options.
          annotations:                   # Annotations defines the annotations of the Ingress object.
            "key": ""
          host: ""                       # Host defines the hostname of the Ingress object.
          ingressClassName: ""           # IngressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource.
          route:                         # Route defines OpenShift Route specific options.
            termination: ""              # Termination specifies the termination type. By default "edge" is used.
          type: ""                       # Type defines the type of Ingress for the Jaeger Query UI. Currently ingress, route and none are supported.
```
(ingressClassName should have been under ingress, as it doesn't apply to route)

Do we value we value _consistency inside the same CR_ or _consistency between the two CRs more_?

internal/manifests/mutate.go
andreasgerstmayr: I did initially, but printing structs with `fmt.Sprintf("%v", some_struct)` is an unreadable mess if the struct is big.

controllers/tempo/tempomonolithic_controller.go
andreasgerstmayr: Good idea, done.

apis/tempo/v1alpha1/tempomonolithic_types.go
rubenvp8510: I would prefer consistency on the same CRD, just one question, in the future is possible to have some sort of consolidation in order to get both? That will imply a breaking change though

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// Observability defines observability configuration for the Tempo deployment
+	//
+	// +kubebuilder:validation:Optional
+	Observability *MonolithicObservabilitySpec `json:"observability,omitempty"`
+
+	// ExtraConfig defines any extra (overlay) configuration for components
+	//
+	// +kubebuilder:validation:Optional
+	ExtraConfig *MonolithicExtraConfigSpec `json:"extraConfig,omitempty"`
```
rubenvp8510, line 39: Any reason for not reusing the same as microservices? The only reason I think is because the microservices could include other configs in the future. If that is the reason I'm ok

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// MonolithicStorageSpec defines the storage for the Tempo deployment.
+type MonolithicStorageSpec struct {
```
rubenvp8510, line 43: One question, Why this have it's own spec and inside only have one structure? Is this for mimic the tempo configuration?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -0,0 +1,244 @@
...
+	// +kubebuilder:default="10Gi"
+	Size resource.Quantity `json:"size"`
+}
+
+// MonolithicIngestionSpec defines the ingestion settings.
+type MonolithicIngestionSpec struct {
+	// OTLP defines the ingestion configuration for OTLP
+	//
+	// +kubebuilder:validation:Optional
+	OTLP *MonolithicIngestionOTLPSpec `json:"otlp,omitempty"`
```
rubenvp8510, line 104: It will be the only protocol supported?

apis/tempo/v1alpha1/tempomonolithic_types.go
rubenvp8510: Same comment, I'd prefer consistency on the same CR

internal/manifests/mutate.go
pavolloffay: shall we then remove fields that are not used in the error struct?

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Yes, it's to mimic the tempo configuration. I thought maybe tempo has plans to store other things in the future, so I'll keep the same here also.

But I don't have very strong opinions on this, I could remove that extra layer if you like.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: I want to make the gateway a drop-in feature, so the service ports should not change if gateway is enabled or not. So I can only support protocols which the gateway also supports.
Afaics the gateway only supports otlp/grpc and otlp/http, right?

With the general move to the OTEL SDK, and using the OTEL collector, I think it's fine to only support OTLP.

internal/manifests/mutate.go
andreasgerstmayr: I updated the message to show the result of `cmp.Diff()` now - it's still a bit unreadable as it's a single line, but when replacing the `\n` and `\t` we see a nice diff in the logs, and we already had a dependency on this library anyway (https://github.com/google/go-cmp).

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: I've updated the PR and reused the struct and logic from the TempoStack now.

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: Yes, I was thinking of maybe doing this change in v1alpha2 of TempoStack if we have a consensus.

pavolloffay (approved):

Issue #721, opened 2023-12-21
Replicas + resources.total.limits


devoxel:
The resource sizing [logic](https://github.com/grafana/tempo-operator/blob/6daca21fa06ea4a13ff0201822e825946471c820/internal/manifests/manifestutils/resources.go#L36) does not take into account the number of replicas assigned.

To demonstrate, here is the following TempoStack definition

```
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoStack
metadata:
  name: simplest
spec:
  images:
    tempo: docker.io/grafana/tempo:x.y.z
    tempoQuery: docker.io/grafana/tempo-query:x.y.z
    tempoGateway: quay.io/observatorium/api
    tempoGatewayOPA: quay.io/observatorium/opa-openshift
  storage:
    secret:
      name: minio-test
      type: s3
  resources:
    total:
      limits:
        memory: 50Gi
        cpu: 10000m
# uncomment and test the difference in output
# template:
#   compactor:
#     replicas: 5
  storageSize: 1Gi
 ```
 
The diff between the version with a comment and without is:

```
<   replicas: 5
---
>   replicas: 1
```

I think this behaviour is not intuitive, and might lead to people over-provisioning their cluster by accident.

andreasgerstmayr:
Thanks for the report! Yep, the assigned limits per component should be divided by the number of replicas for this component.

rafiramadhana:
can I help with this? thx

PR #725, opened 2023-12-30
Divide assigned limits with replicas

rafiramadhana:
Resolves https://github.com/grafana/tempo-operator/issues/721

### DIFF
--- a/internal/manifests/manifestutils/resources.go
+++ b/internal/manifests/manifestutils/resources.go
@@ -49,6 +49,9 @@ func Resources(tempo v1alpha1.TempoStack, component string) corev1.ResourceRequi
 	if ok {
 		totalCpuInt := totalCpu.MilliValue()
 		cpu := float32(totalCpuInt) * componentResources.cpu
+		if r := tempo.Spec.Template.Compactor.Replicas; r != nil && *r > 1 {
+			cpu /= float32(*r)
+		}
 
 		resources.Limits = corev1.ResourceList{
 			corev1.ResourceCPU: *resource.NewMilliQuantity(int64(cpu), resource.BinarySI),
@@ -68,6 +71,9 @@ func Resources(tempo v1alpha1.TempoStack, component string) corev1.ResourceRequi
 		}
 		totalMemoryInt := totalMemory.Value()
 		mem := float32(totalMemoryInt) * componentResources.memory
+		if r := tempo.Spec.Template.Compactor.Replicas; r != nil && *r > 1 {
+			mem /= float32(*r)
+		}
 		resources.Limits[corev1.ResourceMemory] = *resource.NewQuantity(int64(mem), resource.BinarySI)
 		resources.Requests[corev1.ResourceMemory] = *resource.NewQuantity(int64(mem*0.3), resource.BinarySI)
 	}
--- a/internal/manifests/manifestutils/resources.go
+++ b/internal/manifests/manifestutils/resources.go
@@ -33,7 +33,7 @@ var (
 )
 
 // Resources calculates the resource requirements of a specific component.
-func Resources(tempo v1alpha1.TempoStack, component string) corev1.ResourceRequirements {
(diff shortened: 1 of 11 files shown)


rubenvp8510:
This looks good so far! Thanks for your contribution!

The reason CI is failing is because some unit tests regarding of how resources are calculated should be updated

rafiramadhana:
> The reason CI is failing is because some unit tests regarding of how resources are calculated should be updated

ok, will take a look

internal/manifests/manifestutils/resources.go
```diff
@@ -49,6 +49,9 @@ func Resources(tempo v1alpha1.TempoStack, component string) corev1.ResourceRequi
 	if ok {
 		totalCpuInt := totalCpu.MilliValue()
 		cpu := float32(totalCpuInt) * componentResources.cpu
+		if r := tempo.Spec.Template.Compactor.Replicas; r != nil && *r > 1 {
```
andreasgerstmayr, line 52: `Resources()` is called from every tempo component (compactor, distributor, ingester, etc). I'd suggest to add an additional function parameter `replicas *int32` to `Resources()` and update all callers.

andreasgerstmayr (changes requested):

internal/manifests/manifestutils/resources.go
rafiramadhana: > Resources() is called from every tempo component (compactor, distributor, ingester, etc).

I see, but would there be any other possibilities for those components to have a replicas value other than `tempo.Spec.Template.Compactor.Replicas`?

because what I imagined is something like this

```
--- Resources:       manifestutils.Resources(tempo, manifestutils.CompactorComponentName),
+++ Resources:       manifestutils.Resources(tempo, manifestutils.CompactorComponentName, tempo.Spec.Template.Compactor.Replicas)
```

(file:  [compactor.go](https://github.com/grafana/tempo-operator/blob/147be1594663788b337d3ea9327f0543b18160e4/internal/manifests/compactor/compactor.go#L111))

andreasgerstmayr: Yes, each component can have individual replica settings - e.g. you can have 2 replicas of the compactor and 10 of the distributor.

https://github.com/grafana/tempo-operator/blob/505a7ebb0aa70dc84002b171c79b62fae33c7a54/internal/manifests/distributor/distributor.go#L197
https://github.com/grafana/tempo-operator/blob/505a7ebb0aa70dc84002b171c79b62fae33c7a54/internal/manifests/ingester/ingester.go#L66
etc.

rafiramadhana: i see, thanks for the pointers

i will do the update

rafiramadhana: updated, ready to review

internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -231,7 +231,7 @@ func deployment(params manifestutils.Params) (*appsv1.Deployment, error) {
 					MountPath: manifestutils.TmpStoragePath,
 				},
 			},
-			Resources: manifestutils.Resources(tempo, manifestutils.QueryFrontendComponentName),
+			Resources: manifestutils.Resources(tempo, manifestutils.QueryFrontendComponentName, tempo.Spec.Template.QueryFrontend.Replicas),
```
andreasgerstmayr, line 234: @rubenvp8510 looks like we currently assign the query-frontend resources to both the query-frontend and tempo-query containers, so we're using the resources allocated to the query-frontend component twice :open_mouth:

andreasgerstmayr (approved):
Looks good, thank you for your contributions @rafiramadhana! :rocket:

rafiramadhana:
it looks like the test was failing due to intermittent error @rubenvp8510 @andreasgerstmayr

andreasgerstmayr:
> it looks like the test was failing due to intermittent error @rubenvp8510 @andreasgerstmayr

you're right, this was caused by a flaky test (fixed in #728)

This PR is good to merge, thank you!

PR #752, opened 2024-01-18
Run operator hub workflow after images has been published

rubenvp8510:


rubenvp8510:
@andreasgerstmayr this is what we talked about, the only thing is that I think the publish image process run twice, one for the release and other in main. I don't think that is a big deal.

andreasgerstmayr (approved):
lgtm, nice work!
hopefully this will resolve our problems with the operatorhub PRs

PR #757, opened 2024-01-20
Fix nil pointer when OIDC is not specified in static mode

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/647


andreasgerstmayr (approved):

rubenvp8510:
> On a second thought, could you also add a check in `oidc_secret.go` before accessing the pointer?
> 
> I'm wondering if there could be a case where the validating webhook doesn't run. In any case, I think it's good practice to check immediately before accessing any (possibly nil) pointers.

Good catch, what I don't get is why that secret is optional? if it's not specified how the OIDC will work? where to get clientID/clientSecret?

PR #789, opened 2024-02-09
Allow resource limits overrides

### LINKED ISSUE #726: Allow resource limits overrides
Currently, Tempo Stack controls how resources are distributed amongst the Tempo deployments. This might lead to the resource wasting. While some of the deployments have overcommitted limits, the other struggle to get more. Just as in my case:
image

You can clearly see that the Ingester does not need such high resources reserved while the Compactor practically gets OOM Killed all the time. Of course I have scaled up to a bigger instance to keep compactor alive, but then I have tons of resources being wasted. 

How about Tempo Stack allowed overriding components limits? E.g like that:
```
kind: TempoStack
metadata:
  name: stack
  namespace: tempons
spec:
  resources:
    total:
      limits:
        memory: 24Gi
        cpu: 1000m
  template:
    compactor:
      resources:
        limits:
          cpu: 500m
          memory: 16Gi
    queryFrontend:
      resources:
        limits:
          cpu: 100m
          memory: 256Mi
```

rubenvp8510:
Fixes #726 

This is the first step on providing some enhancements to the resources, this at least will provide some flexibility for the users that are not comfortable using the formula.

This doesn't change anything on the formula side, it will directly override the component resources. and won't touch the rest of the components at all.


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -527,6 +527,12 @@ type TempoComponentSpec struct {
 	// +listType=atomic
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tolerations"
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Resources defines resources for this component, this will override the calculated resources derived from total
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resources"
+	Resources *corev1.ResourceRequirements `json:"total,omitempty"`
```
andreasgerstmayr, line 535:
```suggestion
	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```

mncf9:
Thank you for this work 💪🙏

PR #799, opened 2024-02-13
Extend TempoMonolithic validations

andreasgerstmayr:
Extend TempoMonolithic validations

### DIFF
--- a/internal/webhooks/validations.go
+++ b/internal/webhooks/validations.go
@@ -1,0 +1,22 @@
+package webhooks
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+const maxLabelLength = 63
+
+func validateName(name string) field.ErrorList {
+	// We need to check this because the name is used as a label value for app.kubernetes.io/instance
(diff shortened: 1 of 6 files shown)


rubenvp8510 (approved):

internal/webhooks/validations.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+const maxLabelLength = 63
+
+func validateName(name string) field.ErrorList {
+	// We need to check this because the name is used as a label value for app.kubernetes.io/instance
+	// Only validate the length, because the DNS rules are enforced by the functions in the `naming` package.
+	if len(name) > maxLabelLength {
```
rubenvp8510, line 14: One question

Is this enough? There are no names that we use (for example concatenating the component name) that will produce names larger that the allowed?

andreasgerstmayr: I think so, because
* max label value length: 63 (for the labels we never use `tempo-$name-component`, only the name)
* max object name length: 253

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

PR #813, opened 2024-02-21
Add TLS docs

andreasgerstmayr:
Added a document with an overview of our server TLS settings for each pod.

@rubenvp8510 is this doc accurate?

Also, a few questions:
* Should the gateway use the certificate and CA of `spec.template.distributor.tls` if configured?
  * if the gateway is enabled and `spec.template.distributor.tls` is configured, will it break gateway -> distributor connections?
* Currently there is no way to configure TLS for the gateway on non-OpenShift clusters?
* Should the TLS config be more flexible? allow choosing between
  * `internal` (self-signed by the operator)
  * `serving-ca` ([generated by OpenShift](https://docs.openshift.com/container-platform/4.14/security/certificates/service-serving-certificate.html))?
  * `custom` (ConfigMap and Secret)


docs/tls.md
```diff
@@ -0,0 +1,42 @@
...
+8080  | public (HTTP)         | `servingCertsService`           | service-ca-operator | no
+8090  | public (gRPC)         | `servingCertsService`           | service-ca-operator | no
+8081  | internal (HTTP)       | `featureGates.httpEncryption`   | internal            | no
+
+## TLS Clients
+Client | TLS Enabled        | TLS Certificate | Notes
+------ | ------------------ | --------------- | -----
+S3     | `spec.storage.tls` | custom          | only custom CA is supported
+Azure  | no                 | -               |
+GCP    | no                 | -               |
```
iblancasa, line 42: Does this mean TLS cannot be used for those storages?

iblancasa (approved):

docs/tls.md
andreasgerstmayr: Tempo doesn't expose any TLS settings for Azure or GCP: https://grafana.com/docs/tempo/latest/configuration/#storage

When using `https://` for the endpoint it will use TLS, but it'll use the certificate bundle of the container. I'll clarify that.

rubenvp8510:
> Added a document with an overview of our server TLS settings for each pod.
> 
> @rubenvp8510 is this doc accurate?
> 
> Also, a few questions:
> 
>     * Should the gateway use the certificate and CA of `spec.template.distributor.tls` if configured?
>       
>       * if the gateway is enabled and `spec.template.distributor.tls` is configured, will it break gateway -> distributor connections?
I remember we prevent this when we generate the cofniguration, if the gateway is enabled TLS configuration is ignored., actually I think it will generate a bad configuration. https://github.com/grafana/tempo-operator/blob/main/internal/manifests/config/tempo-config.yaml#L12 I'll test today afternoon. I think if that is the case we need to add some validations to the webhook.
> 
>     * Currently there is no way to configure TLS for the gateway on non-OpenShift clusters?
It seems is not :/ 
> 
>     * Should the TLS config be more flexible? allow choosing between
>       
>       * `internal` (self-signed by the operator)
>       * `serving-ca` ([generated by OpenShift](https://docs.openshift.com/container-platform/4.14/security/certificates/service-serving-certificate.html))?
>       * `custom` (ConfigMap and Secret)

I think yes, in the case of self-signed vs custom, you can control this using the feature flags, wondering if we need to control this better on the CRD.

PR #814, opened 2024-02-21
Generalize gateway logic

andreasgerstmayr:
* remove TempoStack parameter (to support TempoMonolithic instances in the future)
* remove hardcoded references in gateway-tenants.yaml template
* move constants to manifestutils/constants.go
* move collecting tenant parameters to `getTenantParams` (shared between TempoStack and TempoMonolithic)


controllers/tempo/common.go
```diff
@@ -118,3 +124,56 @@ func listFieldErrors(fieldErrs field.ErrorList) string {
...
 }
+
+func getTenantParams(
+	ctx context.Context,
+	k8sclient client.Client,
+	ctrlConfig *configv1alpha1.ProjectConfig,
+	namespace string,
+	name string,
+	tenants v1alpha1.TenantsSpec,
+	gatewayEnabled bool,
```
rubenvp8510, line 135: wow, a lot of parameters :(

andreasgerstmayr: I know, do you have a suggestion to improve it?
I think it's nice to have this one function called by `TempoStack` and `TempoMonolithic`, as both need the same logic.

rubenvp8510 (approved):

controllers/tempo/common.go
rubenvp8510: There are no too much options, I think it is fine. the other option is create an structure, but seems to be unnecessary.

PR #815, opened 2024-02-21
Add multitenancy spec to TempoMonolithic CR and update defaults

andreasgerstmayr:
Add multitenancy spec to TempoMonolithic CR and update defaults


apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -277,14 +282,36 @@ type MonolithicJaegerUIRouteSpec struct {
...
 
+// MonolithicMultitenancySpec defines the multi-tenancy settings for Tempo.
+type MonolithicMultitenancySpec struct {
+	// Enabled defines if multi-tenancy is enabled.
+	//
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enabled",xDescriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	Enabled bool `json:"enabled"`
+
+	TenantsSpec `json:",inline"`
```
pavolloffay, line 299: missing docs?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -277,14 +282,36 @@ type MonolithicJaegerUIRouteSpec struct {
...
+
+	// Resources defines the compute resource requirements of the gateway container.
```
pavolloffay, line 301: what is the gateway? Shall we explain it a little bit here?

apis/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -409,7 +436,7 @@ type TempoMonolithicStatus struct {
 
 // TempoMonolithic manages a Tempo deployment in monolithic mode.
 //
-// +operator-sdk:csv:customresourcedefinitions:displayName="TempoMonolithic",resources={{ConfigMap,v1},{Service,v1},{StatefulSet,v1},{Ingress,v1},{Route,v1}}
+// +operator-sdk:csv:customresourcedefinitions:displayName="TempoMonolithic",resources={{ConfigMap,v1},{ServiceAccount,v1},{Service,v1},{Secret,v1},{StatefulSet,v1},{Ingress,v1},{Route,v1}}
```
pavolloffay, line 439: why are we changing this? 

Is it expected that users will interact with these objects? https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/building-your-csv.md#owned-crds

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: It's an inline field, docs are at the properties of the inlined struct

apis/tempo/v1alpha1/tempomonolithic_types.go
andreasgerstmayr: I did put all CSV-related changes in this PR, so the other PR is easier to review without all the generated files.

> Resources: Your CRDs will own one or more types of Kubernetes objects. These are listed in the resources section to inform your end-users of the objects they might need to troubleshoot or how to connect to the application, such as the Service or Ingress rule that exposes a database.

The operator owns a `ServiceAccount` of the deployment, and, if the gateway is enabled, also a `Secret`.

apis/tempo/v1alpha1/common_types.go
```diff
@@ -33,3 +36,13 @@ type TLSSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Min TLS Version"
 	MinVersion string `json:"minVersion,omitempty"`
 }
+
+// ExtraConfigSpec defines extra configurations for tempo that will be merged with the operator generated, configurations defined here
+// has precedence and could override generated config.
+type ExtraConfigSpec struct {
+	// Tempo defines any extra Tempo configuration, which will be merged with the operator's generated Tempo configuration
```
pavolloffay, line 43: we should also document the implications. Is arbitrary configuration supported?

PR #816, opened 2024-02-21
Support multi-tenancy in TempoMonolithic CR

andreasgerstmayr:
Support multi-tenancy in TempoMonolithic CR.


internal/manifests/monolithic/configmap.go
```diff
@@ -152,7 +154,19 @@ func buildTempoConfig(opts Options) ([]byte, error) {
...
+	if tempo.Spec.Multitenancy.IsGatewayEnabled() {
+		// all connections to tempo must go via gateway
+		config.Server.HTTPListenAddress = "localhost"
+		config.Server.GRPCListenAddress = "localhost"
+	}
+
+	// The internal server is required because if the gateway is enabled,
+	// the Tempo API will listen on localhost only,
+	// and then Kubernetes cannot reach the health check endpoint.
+	config.InternalServer.Enable = true
```
rubenvp8510, line 168: Wondering if we cannot have the internal server always enabled? :D

rubenvp8510 (approved):
Small comment, no blocker LGTM!

internal/manifests/monolithic/configmap.go
andreasgerstmayr: Yes, I think we should always enable it in the `TempoStack` as well.
It is always enabled in `TempoMonolithic`.

PR #837, opened 2024-03-07
Add metrics for created operands

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/838

This PR adds two metrics to the operator prometheus endpoints, which indicates how many instances per storage are u sed, and how many instances are managed/unmanaged.


internal/crdmetrics/tempo_stack_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+func TestValueObservedMetrics(t *testing.T) {
+	s := scheme.Scheme
+
+	// Add jaeger to schema
+	s.AddKnownTypes(v1alpha1.GroupVersion, &v1alpha1.TempoStack{}, &v1alpha1.TempoStackList{})
+
+	// Create jaeger instances
```
iblancasa, line 102:
```suggestion
	// Create tempo instances
```

internal/crdmetrics/tempo_stack_test.go
```diff
@@ -0,0 +1,182 @@
...
+		&tempoUnmanaged,
+		&tempoUnmanaged2,
+	}
+	expected := []expectedMetric{
+		newExpectedMetric(managedMetric, attribute.String("state", string(v1alpha1.ManagementStateManaged)), 3),
+		newExpectedMetric(managedMetric, attribute.String("state", string(v1alpha1.ManagementStateUnmanaged)), 2),
+		newExpectedMetric(storageBackendMetric, attribute.String("type", string(v1alpha1.ObjectStorageSecretGCS)), 1),
+		newExpectedMetric(storageBackendMetric, attribute.String("type", string(v1alpha1.ObjectStorageSecretS3)), 2),
+		newExpectedMetric(storageBackendMetric, attribute.String("type", string(v1alpha1.ObjectStorageSecretAzure)), 2),
+	}
```
iblancasa, line 141: Wouldn't it make sense to do this assertions in a loop?

rubenvp8510: At the end this is evaluated in a loop, but this way it makes me easier to see which metrics I'm evaluating. I just fin easier to read.

iblancasa (approved):

internal/crdmetrics/tempo_stack_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/grafana/tempo-operator/apis/tempo/v1alpha1"
+)
+
+const (
+	AgentSideCar   = "Sidecar"
```
andreasgerstmayr, line 23: Is this used anywhere?

rubenvp8510: Actually no. I'll remove it.

andreasgerstmayr (commented):
LGTM :+1:

Could we add the new metrics to the existing `monitoring` e2e test?
I think we also need to set the `openshift.io/cluster-monitoring: "true"` label on the namespace (in the e2e test), but except that, only the `check_metrics.sh` script needs to be update to add the new metric names afaics.

andreasgerstmayr (approved):

rubenvp8510:
> LGTM 👍
> 
> Could we add the new metrics to the existing `monitoring` e2e test? I think we also need to set the `openshift.io/cluster-monitoring: "true"` label on the namespace (in the e2e test), but except that, only the `check_metrics.sh` script needs to be update to add the new metric names afaics.

I'll address this in a separate PR :) but yes, it worth to add it to our e2e.

PR #839, opened 2024-03-07
Add a single e2e job that can be used in branch protection

pavolloffay:


andreasgerstmayr:
Can we rename the jobs to `End-to-end tests (min)` and `End-to-end tests (max)` or some other static job name? Then we don't have to spin up two extra VMs/jobs.

.github/workflows/e2e.yaml
```diff
@@ -114,3 +114,30 @@ jobs:
 
     - name: Run tests
       run: make e2e-upgrade
+  e2e-tests-check:
```
pavolloffay, line 117: @andreasgerstmayr would it mean moving it into `e2e-tests` ? 

Is the number of VMs a problem?

andreasgerstmayr: My understanding is that the full job name, for example `End-to-end tests (1.25)`, is used for the branch protection rules. If we would have a static name, for example `End-to-end tests (min)`, it wouldn't be an issue when we change the min/max version.

So we could have a matrix
```
matrix:
    include:
      - name: min
        kube-version: 1.25
      - name: max
        kube-version: 1.29
```

and use `name: End-to-end tests (${{ matrix.name }})` in the job name.

Or we could also just drop the matrix configuration entirely (as we have only 2 configurations anyway), duplicate the e2e test and give it a static name. We can define the min and max version somewhere globally, e.g.
```
env:
  MIN_KUBERNETES_VERSION: 1.25
  MAX_KUBERNETES_VERSION: 1.29
```

> Is the number of VMs a problem?

They take a while to start, and there's a limit of how many concurrent jobs we have per project (afaics 20). We currently have 9 jobs per commit, I don't see a benefit in adding 2 more for the purpose of the branch protection rules. I'd like to decrease the overall CI time, maybe we can do something to decrease the time of the e2e tests (but that's another topic).

pavolloffay:
Closed in favour of https://github.com/grafana/tempo-operator/pull/843

PR #847, opened 2024-03-12
Monolithic: use internal Tempo server

andreasgerstmayr:
If the gateway is enabled, we want to prevent direct connections to the
Tempo API by making it listen to localhost. All connections must go via
gateway for authentication and authorization.

The internal Tempo server listens on all interfaces and responds to
health checks.

### DIFF
--- a/.chloggen/monolithic_internal_server.yaml
+++ b/.chloggen/monolithic_internal_server.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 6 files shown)


pavolloffay:
@andreasgerstmayr could you please add a changelog and explain why the use-case? Is it for enabling the probe?

andreasgerstmayr:
> @andreasgerstmayr could you please add a changelog and explain why the use-case? Is it for enabling the probe?

Exactly, the internal server is for serving the health checks. If the Tempi API server listens on localhost, no-one outside the pod can access is, also not the service running the Kubernetes health checks.

I've added a changelog entry.

pavolloffay (approved):

PR #850, opened 2024-03-13
Support upgrading TempoMonolithic CR

andreasgerstmayr:
Support upgrading TempoMonolithic CR
* the upgrade function accepts any object implementing `UpgradeableCR`
* the upgrade now only runs in the reconcile loop and not at startup, because the operator reconciles all managed objects at startup anyway
* removed metric series `tempooperator_upgrades_total{state="up-to-date"}` because it was only counted at operator startup, and didn't count any instances created during operator lifetime


rubenvp8510 (approved):
LGTM, only need to resolve conflicts.

PR #877, opened 2024-04-16
Use tempooperatorbot for creating the release PR

rubenvp8510:
This changes a little bit how the things work on the release workflow

  - Now the PR is created by the `tempooperatorbot`, this is because the PRs created by the github actions bot doesn't trigger other actions, by design. So the only way to trigger the CI on the release PR is to create it with other account.
 
 - Now the release action doesn't listen for PR closed, instead it will triggered when a push happens on main, and the `CHANGELOG.md `file changed. This is because due the changes on the first point, the token for the pull_request event is now read-only and doesn't allow us to create the release.
  
  - The version extraction now used `CHANGELOG.md` to extract the latest version and create the tag and release, because we don't use pull_request event anymore, we don't have access to the source branch. Also `CHANGELOG.md`  should contain the latest version. so it is a good candidate for this process.


.github/workflows/release.yaml
```diff
@@ -1,32 +1,27 @@
 name: "Release"
 on:
-  pull_request:
+  push:
     branches:
-      - main
-    types:
-      - closed
+      - 'release/**'
+      - 'main'
```
andreasgerstmayr, line 6:
```suggestion
```

afaics we don't need the release workflow for pushes to main, but only for the `release` branches?

andreasgerstmayr (commented):
LGTM, I left one comment regarding the branch triggers of the `release` workflow

.github/workflows/release.yaml
rubenvp8510: Is this for listen when someone push to main right? this happens when the release PR is merged.

andreasgerstmayr: Ah, yes, you're right. So we should remove the `release/**` then?
The release workflow should run only after a push to `main`, if the `CHANGELOG.md` got modified

rubenvp8510: Make sense? :D

rubenvp8510: Yes, I think we don't needed.

rubenvp8510: Done!

andreasgerstmayr (approved):

PR #892, opened 2024-04-23
WIP - Enable gateway on single tenancy

rubenvp8510:
This will introduce some breaking changes:
 - URL for the jaeger UI will change
 - Now that all is secure we won't be able to use direct the tempo API as a datasource for grafana
 - only support otlp


andreasgerstmayr:
* also we need to create a ClusterRole/ClusterRoleBinding
* service name changes (distributor --> gateway)

PR #910, opened 2024-05-03
Add support for oauth proxy on single tenant

rubenvp8510:
This adds an option for enable oauth-proxy in front of the jaeger UI

It also adds an option to default enable it in case of using an OpenShift route.


apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -42,6 +45,11 @@ type ImagesSpec struct {
 	//
 	// +optional
 	TempoGatewayOpa string `json:"tempoGatewayOpa,omitempty"`
+
+	// OauthProxy defines the oauth proxy image used to protect the jaegerUI on single tenant.
+	//
+	// +optional
+	OauthProxy string `json:"oauthProxy,omitempty"`
```
pavolloffay, line 52: We were also thinking about using kube-rbac-proxy? Is there any cons/pros for using one or the other?

pavolloffay (commented):
We should make the SAR configurable

apis/config/v1alpha1/projectconfig_types.go
rubenvp8510: The thing with kube-rbac-proxy is that is only for token based authentication. It doesn't support the oauth flow we need to protect UI stuff.

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -655,6 +664,13 @@ type IngressSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Route Configuration"
 	Route RouteSpec `json:"route,omitempty"`
+
+	// Security defines the options for the ingress security in single tenant mode.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Ingress Security Configuration"
+	Security IngressSecuritySpec `json:"security,omitempty"`
```
pavolloffay, line 673: is the ingress used for both - ingestion and query? Would it make sense to scope the `security` (I don't like the term) for just the jaeger query?

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -668,6 +684,23 @@ type RouteSpec struct {
...
+// IngressSecuritySpec defines ingress security options.
+type IngressSecuritySpec struct {
+	// Type defines the security type applied to the ingress
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Type"
+	Type IngressSecurityType `json:"type,omitempty"`
+
+	// SAR defines the SAR to be used in the oauth-proxy
```
pavolloffay, line 696: What is the default value?

apis/tempo/v1alpha1/tempostack_types.go
rubenvp8510: Oh you're right, I'll move this to jaegerUI struct. I also accept other suggestions different from security for the name :D

apis/tempo/v1alpha1/tempostack_types.go
rubenvp8510: Is empty, if you don't specify it doesn't add it to the arguments, may be we should have a default one.

rubenvp8510: I added a default to `"{"namespace": "<tempo_stack_namespace>", "resource": "pods", "verb": "get"}`

apis/tempo/v1alpha1/tempostack_types.go
rubenvp8510: Moved!

Makefile
```diff
@@ -151,6 +154,7 @@ run: manifests generate ## Run a controller from your host.
 	RELATED_IMAGE_TEMPO_QUERY=$(TEMPO_QUERY_IMAGE) \
 	RELATED_IMAGE_TEMPO_GATEWAY=$(TEMPO_GATEWAY_IMAGE) \
 	RELATED_IMAGE_TEMPO_GATEWAY_OPA=$(TEMPO_GATEWAY_OPA_IMAGE) \
+	OAUTH_PROXY_IMAGE=$(OAUTH_PROXY_IMAGE) \
```
andreasgerstmayr, line 157:
```suggestion
	RELATED_IMAGE_OAUTH_PROXY_IMAGE=$(OAUTH_PROXY_IMAGE) \
```

Makefile
```diff
@@ -363,7 +367,7 @@ prepare-e2e: chainsaw start-kind cert-manager set-test-image-vars build docker-b
 
 .PHONY: e2e
 e2e:
-	$(CHAINSAW) test --test-dir ./tests/e2e
+	$(CHAINSAW) test --test-dir ./tests/e2e/compatibility
```
andreasgerstmayr, line 370:
```suggestion
	$(CHAINSAW) test --test-dir ./tests/e2e
```

apis/tempo/v1alpha1/tempostack_types.go
pavolloffay: We should document what the default is

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -13,6 +13,9 @@ import (
 // +kubebuilder:validation:Enum=Managed;Unmanaged
 type ManagementStateType string
 
+// IngressSecurityType defines the type of security for the ingress/route.
+type IngressSecurityType string
```
pavolloffay, line 17: not needed anymore

apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -597,6 +600,13 @@ type JaegerQuerySpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="ServicesQueryDuration"
 	ServicesQueryDuration *metav1.Duration `json:"servicesQueryDuration,omitempty"`
+
+	// Oauth defines the options for the oauth proxy used to protect jaeger UI
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Ingress Security Configuration"
+	Oauth OauthSpec `json:"security,omitempty"`
```
pavolloffay, line 609: Is the route created for query-frontend when the jaeger-ui is disabled?

pavolloffay (commented):
@rubenvp8510 is this feature relevant to the monolithic deployment? 

If so do we want to extend this PR or do it separately?

internal/manifests/queryfrontend/oauth_proxy.go
```diff
@@ -0,0 +1,202 @@
...
+		Env:       proxy.ReadProxyVarsFromEnv(),
+		ReadinessProbe: &corev1.Probe{
+			ProbeHandler: corev1.ProbeHandler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Scheme: corev1.URISchemeHTTPS,
+					Path:   healthPath,
+					Port:   intstr.FromString(manifestutils.OAuthProxyPortName),
+				},
+			},
+			InitialDelaySeconds: 15,
```
pavolloffay, line 173: I would move these two to constants

docs/operator/config.yaml
```diff
@@ -143,6 +143,9 @@ featureGates:
     # If empty the operator automatically derives the domain from the cluster.
     baseDomain: ""
 
+    # OAuthProxyEnabled is used internally for enable by default the oauth proxy for the UI in single tenant mode
+    oAuthProxyEnabled: false
```
andreasgerstmayr, line 147: do we ever need to add more config options related to oauth proxy? if so, I'd make it a block (struct) like
```
oauthproxy:
  enabled: false
```

docs/spec/tempo.grafana.com_tempostacks.yaml
```diff
@@ -182,6 +183,9 @@ spec:                                    # TempoStackSpec defines the desired st
         monitorTab:                      # MonitorTab defines the monitor tab configuration.
           enabled: false                 # Enabled enables the monitor tab in the Jaeger console. The PrometheusEndpoint must be configured to enable this feature.
           prometheusEndpoint: ""         # PrometheusEndpoint defines the endpoint to the Prometheus instance that contains the span rate, error, and duration (RED) metrics. For instance on OpenShift this is set to https://thanos-querier.openshift-monitoring.svc.cluster.local:9091
+        security:                        # Oauth defines the options for the oauth proxy used to protect jaeger UI
```
andreasgerstmayr, line 186: I wonder if we should use a more descriptive name than `security`. Maybe `authentication`?

internal/manifests/queryfrontend/oauth_proxy.go
```diff
@@ -0,0 +1,202 @@
...
+			MountPath: tlsProxyPath,
+			Name:      getTLSSecretNameForFrontendService(tempo),
+		},
+
+			{
+				MountPath: oauthProxySecretMountPath,
+				Name:      cookieSecretName(tempo),
+			},
+		},
+		Resources: resources(tempo),
```
andreasgerstmayr, line 163: That will use the resources defined for `tempo.Spec.Template.QueryFrontend.Resources`?

internal/manifests/queryfrontend/oauth_proxy.go
```diff
@@ -0,0 +1,202 @@
...
+			ProbeHandler: corev1.ProbeHandler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Scheme: corev1.URISchemeHTTPS,
+					Path:   healthPath,
+					Port:   intstr.FromString(manifestutils.OAuthProxyPortName),
+				},
+			},
+			InitialDelaySeconds: 15,
+			TimeoutSeconds:      5,
+		},
```
andreasgerstmayr, line 175: Can you add
```
SecurityContext: manifestutils.TempoContainerSecurityContext(),
```
to the pod template? For enhanced security, and iirc we got warnings in the OpenShift logs otherwise.

internal/manifests/queryfrontend/oauth_proxy.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func oauthProxyService(tempo v1alpha1.TempoStack) *corev1.Service {
```
andreasgerstmayr, line 179: Is a new service for the oauth proxy required? Isn't the proxy just a sidecar in the query-frontend pod?

If it's just for the `service.beta.openshift.io/serving-cert-secret-name`, I'd add this annotation to the query-frontend service.

internal/manifests/queryfrontend/oauth_proxy.go
rubenvp8510: For now yes :/ is the same case for jaegerUI/tempo-querier

docs/spec/tempo.grafana.com_tempostacks.yaml
rubenvp8510: Changed to authenticate!

internal/manifests/queryfrontend/oauth_proxy.go
rubenvp8510: +1 I use the same service, also the port changed to use the oauth-proxy

internal/webhooks/tempostack_webhook.go
```diff
@@ -149,6 +149,20 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
 		r.Spec.Template.QueryFrontend.JaegerQuery.ServicesQueryDuration = &defaultServicesDuration
 	}
 
+	if d.ctrlConfig.Gates.OpenShift.OauthProxy.DefaultEnabled {
+		if r.Spec.Template.QueryFrontend.JaegerQuery.Enabled && r.Spec.Template.QueryFrontend.JaegerQuery.Ingress.Type == v1alpha1.IngressTypeRoute {
+			r.Spec.Template.QueryFrontend.JaegerQuery.Oauth.Enabled = true
```
andreasgerstmayr, line 154: If the user explicitly sets `r.Spec.Template.QueryFrontend.JaegerQuery.Oauth.Enabled` to false, the current code overwrites it every time the webhook runs (if `DefaultEnabled` = true).

So it's not a default anymore, but more a `ForceEnabled` :smile:

Possible solutions
* make the bool a pointer, then we can differentiate between unset, true and false
* remove this setting from the CR, i.e. don't make it user-configurable per tempo instance

rubenvp8510: True! I'll fix this.

rubenvp8510: Fixed!

andreasgerstmayr (approved):

PR #958, opened 2024-06-28
Update tempo to 2.4.2

pavolloffay:


rubenvp8510:
> The ownership of `/var/tempo` changed, see the breaking changes of the 2.5.0 release: https://github.com/grafana/tempo/releases/tag/v2.5.0
> 
> > Ownership of /var/tempo is changing. Historically this has been owned by root:root, and with this change it will now be owned by tempo:tempo with the UID/GID of 10001. The ingester and metrics-generator statefulsets may need to be chown'd in order to start properly. A jsonnet example of an init container is included with the PR. This impacts impacts all users of the grafana/tempo Docker image.
> 
> I think we'll need an init container to adjust existing installations.



> The ownership of `/var/tempo` changed, see the breaking changes of the 2.5.0 release: https://github.com/grafana/tempo/releases/tag/v2.5.0
> 
> > Ownership of /var/tempo is changing. Historically this has been owned by root:root, and with this change it will now be owned by tempo:tempo with the UID/GID of 10001. The ingester and metrics-generator statefulsets may need to be chown'd in order to start properly. A jsonnet example of an init container is included with the PR. This impacts impacts all users of the grafana/tempo Docker image.
> 
> I think we'll need an init container to adjust existing installations.

Is this not something solved by the securityContext?

pavolloffay:
good that we have the upgrade tests

an instance created with 2.4.2 and upgraded to 2.5.0 throws 
```
level=error ts=2024-06-28T14:45:59.767604772Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="rpc error: code = Unknown desc = mkdir /var/tempo/wal/d3ff7b03-3e2e-4380-aa94-d8930e2f97c2+single-tenant+vParquet3: permission denied"
level=error ts=2024-06-28T14:46:00.761519343Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="rpc error: code = Unknown desc = mkdir /var/tempo/wal/82e00424-ff58-4a66-b83b-ecd8f4cdb8c1+single-tenant+vParquet3: permission denied"
level=error ts=2024-06-28T14:46:01.763387853Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="rpc error: code = Unknown desc = mkdir /var/tempo/wal/2bf4ee9e-9701-43b7-858e-f469229ec921+single-tenant+vParquet3: permission denied"
level=error ts=2024-06-28T14:46:02.76407992Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="rpc error: code = Unknown desc = mkdir /var/tempo/wal/a9b65782-2bc0-482c-a185-84eb36cd7301+single-tenant+vParquet3: permission denied"
```

pavolloffay:
PR in the tempo helm chart https://github.com/grafana/helm-charts/pull/3161#issuecomment-2197202507

pavolloffay:
The issue does not seem to occur on OCP. The OCP seems to be setting this `securityContext`  by default

```
  securityContext:
    fsGroup: 1001020000
    seLinuxOptions:
      level: s0:c32,c14
    seccompProfile:
      type: RuntimeDefault
```


Tested with 
```yaml
kubectl apply -f - <<EOF
kind: TempoStack
metadata:
  name: simplest
spec:
  managementState: Managed
  images:
    tempo: docker.io/grafana/tempo:2.4.1
  storage:
    secret:
      name: minio-test
      type: s3
  storageSize: 1Gi
  resources:
    total:
      limits:
        memory: 2Gi
        cpu: 2000m
  template:
    queryFrontend:
      jaegerQuery:
        enabled: true
EOF
```

reported some data and then changed to `docker.io/grafana/tempo:2.5.0`

internal/upgrade/v0_11_0.go
```diff
@@ -0,0 +1,122 @@
...
+	err := u.Client.List(ctx, pvcs, listOps...)
+	if err != nil {
+		return err
+	}
+	return upgrade0_11_0_pvcs(ctx, u, tempo.ObjectMeta, tempo.Spec.NodeSelector, u.CtrlConfig.DefaultImages.Tempo, pvcs)
+}
+
+func upgrade0_11_0_pvcs(ctx context.Context, u Upgrade, tempo metav1.ObjectMeta, nodeSelector map[string]string, image string, pvcs *corev1.PersistentVolumeClaimList) error {
+	// keep the jobs around for 1 day
+	ttl := int32(60 * 60 * 24)
```
rubenvp8510, line 67: Is this TTL necessary?

rubenvp8510 (approved):
Small comment but other than that LGTM

internal/upgrade/v0_11_0.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Upstream Tempo 2.5.0 has a breaking change https://github.com/grafana/tempo/releases/tag/v2.5.0
+// The /var/tempo is created in the dockerfile with 10001:10001
+// The user is changed to 10001:10001
+// The previous user in 2.4.2 was root (0)
+// The Red Hat Tempo image does not use root user (it uses 1001) and on OpenShift the /var/tempo PV has a different fsGroup
+// so the issue does not happen on OpenShift.
+func upgrade0_11_0(ctx context.Context, u Upgrade, tempo *v1alpha1.TempoStack) error {
+	// do nothing on OpenShift
+	if u.CtrlConfig.Gates.OpenShift.OpenShiftRoute {
```
rubenvp8510, line 27: funny way of detect openshift :D

internal/upgrade/v0_11_0.go
pavolloffay: otherwise the job would hang there forever

Dockerfile
```diff
@@ -18,6 +18,7 @@ COPY . .
 
 # Build
 ARG OPERATOR_VERSION
+ARG TEMPO_VERSION
```
pavolloffay, line 21: @rubenvp8510 do you know how `OPERATOR_VERSION` is passed in GHA when the image is published? We might need to update the workflow to include the tempo version

Makefile
```diff
@@ -159,7 +159,7 @@ run: manifests generate ## Run a controller from your host.
 
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	docker buildx build --load --platform linux/${ARCH} --build-arg OPERATOR_VERSION -t ${IMG} .
+	docker buildx build --load --platform linux/${ARCH} --build-arg OPERATOR_VERSION --build-arg TEMPO_VERSION -t ${IMG} .
```
andreasgerstmayr, line 162: afaics we don't need to pass `OPERATOR_VERSION` or `TEMPO_VERSION` here, because in the Dockerfile we run `make build`, and the Makefile contains the versions (hardcoded).

andreasgerstmayr (commented):
I'm wondering about race conditions, this is the timeline on upgrade:

1. old tempo version runs
2. operator restarts with new operator version
3. operator reconciles all managed instances
3.1. at the start of reconcile loop, operator runs upgrades
3.2. upgrade procedure creates job to `chown` PV contents **(while old tempo pods are still running)**
3.3. reconcile create the manifests, using the new tempo image in all `Deployment` and `StatefulSet` manifests
3.4. reconcile detects changes and applies the new manifests to the cluster, triggering a restart of the tempo pods **(the upgrade job may or may not be completed at this time)**

I guess this works in most cases, but not all (busy cluster, chown-job takes long time to schedule/execute, the ingester keeps writing during chmod, so we end up with some root-owned and some tempo-owned files in the PV, etc).

Ideally we stop running ingesters, perform `chown`, wait until `chown` is done, and then only start the new version.

I think adding an initContainer would solve the timing issues. init containers always start and finish before the main container, and afaics (correct me if I'm wrong) while the init container of the new statefulset runs, the old statefulset is already terminated (otherwise two pods access the same PV, I think this is prevented).

pavolloffay:
> I think adding an initContainer would solve the timing issues. init containers always start and finish before the main container, and afaics (correct me if I'm wrong) while the init container of the new statefulset runs, the old statefulset is already terminated (otherwise two pods access the same PV, I think this is prevented).

There is an issue with the init-container approach. The init-container will run only for ingesters (and PVs) that are scheduled to run. The ingesters might be scaled down (e.g. from 5 instances to 2) therefore there will be PVs that won't be upgraded.

Makefile
andreasgerstmayr: Ah, yes, but this is only required for testing, right? Anyway, still useful :+1:

pavolloffay:
> We'll have to keep the initContainer indefinitely 

This is certainly doable but it will clutter non upgrade related packages. I would prefer to keep the upgrade code in an isolated standalone package.

I have changed the upgrade procedure to scale ingesters down to zero before upgrading, then wait until a single job that mounts all volumes finishes.

internal/upgrade/v0_11_0.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			return false, err
+		}
+		if job.Status.Succeeded == 1 {
+			return true, nil
+		}
+
+		return false, nil
+	})
+
```
andreasgerstmayr, line 197: we could remove the job here, instead of leaving it there until the 24h TTL. but it's fine either way.

andreasgerstmayr (approved):
lgtm, nice job (pun intended)

internal/upgrade/v0_11_0.go
pavolloffay: I left it there intentionally for awareness :)

PR #962, opened 2024-07-03
Use TLS via OpenShift service annotation when gateway/multitenancy is…

rubenvp8510:
… disabled

Fixes https://github.com/grafana/tempo-operator/issues/963

This PR doesn't cover monolitic. I would prefer to do it in a separate PR. in this way I can keep this PR small and move forward faster.


apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -95,6 +95,9 @@ type OpenShiftFeatureGates struct {
 
 	// OauthProxy define options for the oauth proxy feature.
 	OauthProxy OauthProxyFeatureGates `json:"oAuthProxy,omitempty"`
+
+	// DistributorServingCertsService
```
pavolloffay, line 99: A more explanation would be helpful? 

What does the config enable? Does it inject OCP service CA to distributor pod? How is it handled for the gateway?

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use TLS via OpenShift service annotation when gateway/multitenancy is disabled
```
pavolloffay, line 8: It  says gateway but I don't see code chagnes for gateway.

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# One or more tracking issues related to the change
+issues: [963]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
pavolloffay, line 16: Could you please document how users could setup TLS to report data with the OCP service CA?

Is it enabled by default on OCP?

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: it says when the gateway/multitenancy is disabled :)

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: I will include a brief explanation here. I think a more elaborate way on how to configure this on the client side (may be using otel collector) should be in the documentation and not in the subtext . Just my opinion.

rubenvp8510: Added a brief explanation, as I said I would prefer a more deep explanation on the documentation. As this is only a changelog. But if you thing I should include something here I'm not opposites.

rubenvp8510:
@pavolloffay Should we make TLS enable by default on OpenShift? following the principle of "making secure by default"?

pavolloffay:
Do you mean using servicing certs on the Gateway public HTTTP and gRPC APIs (the users would need to use the service  CA)?

rubenvp8510:
> Do you mean using servicing certs on the Gateway public HTTTP and gRPC APIs (the users would need to use the service CA)?

I don't understand well the question. From what I can see the gateway case is already implemented. This PR what it does is to use the service CA for the case when the gateway is not enabled. This means configure the distributor to use it directly (without the gateway)

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,27 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This will use the openshift service annotation to generate TLS certs, in order to use this feature TLS on the 
+  distributor should be enabled in this way:
+  ```
+    apiVersion: tempo.grafana.com/v1alpha1
+    kind:  TempoStack
+    spec:
+        distributor:
```
pavolloffay, line 23: there is no `distributor` under `spec` .

apis/config/v1alpha1/projectconfig_types.go
```diff
@@ -95,6 +95,12 @@ type OpenShiftFeatureGates struct {
 
 	// OauthProxy define options for the oauth proxy feature.
 	OauthProxy OauthProxyFeatureGates `json:"oAuthProxy,omitempty"`
+
+	// DistributorServingCertsService enables OpenShift service-ca annotations on the Distributor service
+	// to use the in-platform CA and generate a TLS cert/key pair per service for in-cluster data-in-transit encryption.
+	// to use this, you need to enable TLS on the distributor and leave caName and certName empty. If you specify caName
+	// and certName, those will be used instead of the ones generated by OpenShift.
+	DistributorServingCertsService bool `json:"distributorServingCertsService,omitempty"`
```
andreasgerstmayr, line 103: I'm wondering if this should be a global setting, applying to all TempoStack and TempoMonolithic instances, or instead a setting on the `TLSSpec`:
```
ServingCert bool `json:"servingCert,omitempty"`
```

I think it would be awesome if, for every TLS configuration we support in the operator, the user can choose between disabling tls, using the servingcerts service to generate it or using a custom tls secret+configmap.

This will make it available in all places where TLSSpec is used however (but it's not supported everywhere yet), so we should throw an error in the validating webhook if it's used in a place where it's not supported yet.

internal/manifests/distributor/distributor.go
```diff
@@ -338,3 +357,30 @@ func service(tempo v1alpha1.TempoStack) *corev1.Service {
...
+	return fmt.Sprintf("%s-distributor-ca", tempoName)
+}
+
+func getConfigmapCABundle(tempo v1alpha1.TempoStack) *corev1.ConfigMap {
+	name := getTLSConfigMapName(tempo.Name)
+	annotations := map[string]string{
+		"service.beta.openshift.io/inject-cabundle": "true",
+	}
+
+	return &corev1.ConfigMap{
```
andreasgerstmayr, line 374: We use this in a few places, you can use `tls.NewConfigMapCABundle()` to generate this configmap.

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,27 @@
...
+note: Use TLS via OpenShift service annotation when gateway/multitenancy is disabled
+
+# One or more tracking issues related to the change
+issues: [963]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This will use the openshift service annotation to generate TLS certs, in order to use this feature TLS on the
```
pavolloffay, line 17: I would rephrase this. 

`On OpenShift when operator config `DistributorServingCertsService` is enabled and the following TempoStack CR is used. The operator provisions OpenShift serving certificates for the distributor ingest APIs`

I would as well document what users need to do to configure TLS on their clients.

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: my bad, I delete the template part

internal/manifests/distributor/distributor.go
```diff
@@ -338,3 +357,30 @@ func service(tempo v1alpha1.TempoStack) *corev1.Service {
 		},
 	}
 }
+
+func getTLSSecretName(tempoName string) string {
```
andreasgerstmayr, line 361: let's use `naming.ServingCertName()` here

internal/manifests/distributor/distributor.go
```diff
@@ -338,3 +357,30 @@ func service(tempo v1alpha1.TempoStack) *corev1.Service {
...
+	return fmt.Sprintf("%s-distributor-tls", tempoName)
+}
+func getTLSConfigMapName(tempoName string) string {
```
andreasgerstmayr, line 364: let's use `naming.ServingCABundleName()` here

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: Fixed.

apis/config/v1alpha1/projectconfig_types.go
rubenvp8510: I think that would be awesome sure. I can rename it but with the clarification that for now the only supported components will be the distributor and the gateway. And then in a follow up PR I can extend this.

andreasgerstmayr: Yes, sounds good! Afaics in the gateway it's also only half-supported, i.e. for the gateway it currently always uses the serving cert but doesn't support custom certs.

rubenvp8510: Yep, those are the gaps we need to fill and that is why I would like to do in a separate PR :) if not. this PR will be big

internal/manifests/distributor/distributor.go
rubenvp8510: Done

internal/manifests/distributor/distributor.go
rubenvp8510: Done

internal/manifests/distributor/distributor.go
rubenvp8510: Done

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,28 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use TLS via OpenShift service annotation when gateway/multitenancy is disabled
+
+# One or more tracking issues related to the change
+issues: [963]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
pavolloffay, line 16: Could we document here how clients (e.g. OTEL col) can configure TLS? e.g. use `service.beta.openshift.io/inject-cabundle` to provision the cert into a config map or use `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` ?

pavolloffay (approved):
LGTM, 

I would just document into release notes how the clients can inject CA cert.

.chloggen/ingest_tls_openshift.yaml
```diff
@@ -0,0 +1,28 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: operator
```
pavolloffay, line 5: Could you please rebase and change this?

pavolloffay (changes requested):

pavolloffay:
How are we going to expose this in the monolithic CR? It would be great to come up with a similar config. Do you have CRD draft?

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: Sure

rubenvp8510:
> How are we going to expose this in the monolithic CR? It would be great to come up with a similar config. Do you have CRD draft?

I don't have a draft, but I would say we can apply the same, if TLS is enabled but not certName is specified AND we are on openshift. we can use the service Certificate.

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: Done.

.chloggen/ingest_tls_openshift.yaml
rubenvp8510: Done

pavolloffay (approved):

PR #975, opened 2024-07-09
Use openshift sertvice annotation when gateway is disabled (monolithic)

rubenvp8510:
This is the monolithic version 

Depends on https://github.com/grafana/tempo-operator/pull/977

### DIFF
--- a/.chloggen/tls_cert_serv_mono.yaml
+++ b/.chloggen/tls_cert_serv_mono.yaml
@@ -1,0 +1,41 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempomonolithic
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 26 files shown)


.chloggen/tls_cert_serv_mono.yaml
```diff
@@ -0,0 +1,41 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  On OpenShift when operator config `servingCertsService` is enabled and the following TempoMonolithic CR is used.
+  The operator provisions OpenShift serving certificates for the distributor ingest APIs
+
+  ```
+    apiVersion: tempo.grafana.com/v1alpha1
+    kind:  TempoStack
```
pavolloffay, line 22: This does not look like monolithic CR

.chloggen/tls_cert_serv_mono.yaml
```diff
@@ -0,0 +1,41 @@
...
+    spec:
+      ingestion:
+        otlp:
+          grpc:
+            tls:
+              enabled: true
```
pavolloffay, line 28: Could we document this in directly in the CRD as well (for both CRDs)

pavolloffay (approved):

Issue #976, opened 2024-07-10
Invalid statefulset when monolithic is deployed with receiver TLS enable both protocols


rubenvp8510:
In monolithic deployment, if we enable TLS on the OTLP- GRPC and OTLP-HTTP protocols the statefulset generated by the operator is invalid and cannot be ready due this error:

```
  Type     Reason        Age                 From                    Message
  ----     ------        ----                ----                    -------
  Warning  FailedCreate  29s (x12 over 40s)  statefulset-controller  create Pod tempo-simplest-0 in StatefulSet tempo-simplest failed error: Pod "tempo-simplest-0" is invalid: [spec.volumes[4].name: Duplicate value: "custom-ca", spec.volumes[5].name: Duplicate value: "custom-cert", spec.containers[0].volumeMounts[4].mountPath: Invalid value: "/var/run/ca-receiver": must be unique, spec.containers[0].volumeMounts[5].mountPath: Invalid value: "/var/run/tls/receiver": must be unique]

```

CR used:

```
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoMonolithic
metadata:
  name: simplest
spec:
  jaegerui:
    enabled: true
  ingestion:
    otlp:
      grpc:
        enabled: true
        tls:
          enabled: true
          caName: custom-ca
          certName: custom-cert
      http:
        enabled: true
        tls:
          enabled: true
          caName: custom-ca
          certName: custom-cert
```

PR #977, opened 2024-07-10
Allow otlp grpc and http to be enabled with TLS on monolithic

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/976


internal/manifests/monolithic/configmap_test.go
```diff
@@ -222,9 +222,9 @@ distributor:
       protocols:
         grpc:
           tls:
-            client_ca_file: /var/run/ca-receiver/service-ca.crt
-            cert_file: /var/run/tls/receiver/tls.crt
-            key_file: /var/run/tls/receiver/tls.key
+            client_ca_file: /var/run/ca-receiver/grpc/service-ca.crt
```
pavolloffay, line 225: Shall we add a test to use the same certs for `otlp.protocols.http` ?

internal/manifests/manifestutils/tls.go
```diff
@@ -117,3 +125,13 @@ func findContainerIndex(pod *corev1.PodSpec, containerName string) (int, error)
 
 	return -1, fmt.Errorf("cannot find container %s", containerName)
 }
+
+func findVolumeIndex(pod *corev1.PodSpec, volumeName string) (int, error) {
```
pavolloffay, line 129: what is the point of returning error if it is always ignored? It seems like the function could be renamed to `containsVolume(pod, volumeName) bool`

rubenvp8510: Done

internal/manifests/monolithic/configmap_test.go
rubenvp8510: Done

internal/manifests/manifestutils/tls.go
```diff
@@ -27,16 +27,20 @@ func MountCAConfigMap(
...
 	})
-	pod.Volumes = append(pod.Volumes, corev1.Volume{
-		Name: caConfigMap,
-		VolumeSource: corev1.VolumeSource{
-			ConfigMap: &corev1.ConfigMapVolumeSource{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: caConfigMap,
+
+	hasVolume := containsVolume(pod, caConfigMap)
+	if !hasVolume {
```
iblancasa, line 32: How about?
```suggestion
	if !containsVolume(pod, caConfigMap) {
```

internal/manifests/manifestutils/paths.go
```diff
@@ -14,6 +14,16 @@ const (
...
 
+	// ReceiverGRPCTLSCADir is the path that is mounted from the configmap for TLS for receiver.
+	ReceiverGRPCTLSCADir = "/var/run/ca-receiver/grpc"
+	// ReceiverGRPCTLSCertDir returns the mount path of the receivers certificates (for ingesting traces).
+	ReceiverGRPCTLSCertDir = TLSDir + "/receiver/grpc"
+
+	// ReceiverHTTPTLSCADir is the path that is mounted from the configmap for TLS for receiver.
+	ReceiverHTTPTLSCADir = "/var/run/ca-receiver/http"
+	// ReceiverHTTPTLSCertDir returns the mount path of the receivers certificates (for ingesting traces).
+	ReceiverHTTPTLSCertDir = TLSDir + "/receiver/http"
```
iblancasa, line 25: Wht don't we use `path.Join` here?

iblancasa (approved):

internal/manifests/manifestutils/paths.go
rubenvp8510: We can but I don't think we need it, this are constants, we don't need to deal with double slashes, empty strings etc..

internal/manifests/manifestutils/tls.go
rubenvp8510: it's fine. I can do the change

rubenvp8510: Changed! thanks for the review!

pavolloffay (approved):

PR #982, opened 2024-07-12
Sanitize volume names for ca bundle and certificates

rubenvp8510:


pavolloffay (approved):

.chloggen/fix_ca_cofigmap_dots.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: operator
```
pavolloffay, line 5: @rubenvp8510 there is no operator component anymore

rubenvp8510: This is a fix that impacts both, monolithic and tempostack. What component should I put here?

rubenvp8510: that could work :D

PR #995, opened 2024-07-29
Add tempor recording rules to OpenShift bundle

rubenvp8510:


pavolloffay:
@rubenvp8510 could you please describe why is this needed?

PR #1005, opened 2024-08-08
Increase e2e assert timeout

andreasgerstmayr:
We run the e2e tests in parallel, i.e. the more tests we add, the more congestion we get when run in (test) clusters with constrained resources such as GitHub actions.


rubenvp8510:
More time didn't do the trick :(

andreasgerstmayr:
> More time didn't do the trick :(

I think the problem was the exponential back-off delay of the Kubernetes Job scheduler, so more time didn't help because the scheduler was just waiting in the back-off delay before starting a new Pod. I changed the test now from a Job to a Pod with a simple infinite loop (chainsaw will terminate it anyway after the assert timeout), let's see if this works better :)

rubenvp8510 (approved):

PR #1007, opened 2024-08-09
[Chore] Fix monolithic-receivers-tls test

IshwarKanse:


tests/e2e/monolithic-receivers-tls/02-assert.yaml
```diff
@@ -36,3 +36,18 @@ spec:
...
+  labels:
+    app.kubernetes.io/instance: opentelemetry
+    app.kubernetes.io/name: otelcol
+    component: standalone-collector
+status:
+  containerStatuses:
+  - name: opentelemetry-collector
+    ready: true
+    started: true
+  phase: Running
```
andreasgerstmayr, line 53: Shouldn't
```
status:
  readyReplicas: 1
```
of the deployment make sure that the readiness probe of the pod passes?
See https://github.com/kubernetes/kubernetes/issues/100192#issuecomment-797864113

IshwarKanse: Yes, it should for a pod with a single container. I'm just checking few thing out if it helps. I also added a check to see if the Monolithic instance is ready. Lets see if it helps.

IshwarKanse: @andreasgerstmayr It failed for the min K8s version. I'll test once on OCP 4.12 which is the min version we support currently to make sure there is no issue. Looking at the logs, Tempo was ready along with OTEL.

andreasgerstmayr (approved):

PR #1019, opened 2024-08-27
Cleanup operator instance metrics when instance is deleted

pavolloffay:
https://issues.redhat.com/browse/TRACING-4609

Tested with 
```
kubectl apply -f - <<EOF
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoMonolithic
metadata:
  name: simplestmono
spec:
  storage:
    traces:
      backend: s3 
      size: 3Gi 
      s3: 
        secret: minio-test
EOF


kubectl apply -f - <<EOF
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoStack
metadata:
  name: simplest
spec:
  storage:
    secret:
      name: minio-test
      type: s3
  storageSize: 1Gi
EOF
```

Then port-forward the operator 8080 and check for metric `controller_runtime_reconcile_total`

PR #1024, opened 2024-08-29
Fix clearing instance metrics when namespace is deleted

pavolloffay:
Related to #1019

### DIFF
--- a/controllers/tempo/tempostack_controller.go
+++ b/controllers/tempo/tempostack_controller.go
@@ -82,6 +82,8 @@ func (r *TempoStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 			log.Error(err, "unable to fetch TempoStack")
 			return ctrl.Result{}, fmt.Errorf("could not fetch tempo: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


iblancasa (approved):

IshwarKanse:
@pavolloffay Verified the fix with the steps from the jira issue. We are good to merge the PR.

controllers/tempo/tempostack_controller.go
```diff
@@ -82,6 +82,8 @@ func (r *TempoStackReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			log.Error(err, "unable to fetch TempoStack")
 			return ctrl.Result{}, fmt.Errorf("could not fetch tempo: %w", err)
 		}
+		// instance is not found, metrics can be cleared
+		status.ClearTempoStackMetrics(req.Namespace, req.Name)
```
andreasgerstmayr, line 86: do we always get a reconcile event on deletion? could we remove the `ClearTempoStackMetrics()` from the deletion webhook then?

PR #1026, opened 2024-09-06
Change default route to reencrypt for gateway

rubenvp8510:


internal/webhooks/tempostack_webhook.go
```diff
@@ -39,7 +39,7 @@ var (
 type TempoStackWebhook struct {
 }
 
-const defaultRouteGatewayTLSTermination = v1alpha1.TLSRouteTerminationTypePassthrough
+const defaultRouteGatewayTLSTermination = v1alpha1.TLSRouteTerminationTypeReencrypt
```
frzifus, line 42: Do we need a changelog for this?

rubenvp8510: I think so, I mean is an small change, but may be the users wanted to be aware?

.chloggen/reencrypt_gw_default.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
frzifus, line 2: Is it maybe breaking? Since some certs will change.

frzifus (approved):
I am unsure with the change type, but aside from that lgtm!

frzifus:
Ahh, monolithic-receivers-tls failed

rubenvp8510:
> Ahh, monolithic-receivers-tls failed

Flaky test

.chloggen/reencrypt_gw_default.yaml
rubenvp8510: I don't think this is a breaking change, as the other things will remaing working in the same way. I think this only adds a couple of hosts to the cert.

andreasgerstmayr (approved):

PR #1029, opened 2024-09-10
Expose pod security context per component

rubenvp8510:


apis/tempo/v1alpha1/tempostack_types.go
```diff
@@ -523,6 +523,12 @@ type TempoComponentSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resources"
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// SecurityContext defines security context will be applied to all pods of this component.
+	//
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="SecurityContext"
+	SecurityContext *corev1.PodSecurityContext `json:"SecurityContext,omitempty"`
```
andreasgerstmayr, line 531:
```suggestion
	SecurityContext *corev1.PodSecurityContext `json:"securityContext,omitempty"`
```

andreasgerstmayr (commented):
I think we should also set it on the compactor component

rubenvp8510:
@andreasgerstmayr do you think we need to put a default value?

andreasgerstmayr:
> @andreasgerstmayr do you think we need to put a default value?

We do set a default security context on the containers, for example here: https://github.com/grafana/tempo-operator/blob/c39140e6c2b2351ce1baffb88c1cf06a6ad93445/internal/manifests/ingester/ingester.go#L128
It has some common fields, but is not the same type. I'm not sure if we should set any defaults (`RunAsNonRoot` maybe? but that could break if users use a custom Tempo image).

Do you think the container security context should also be customizable? Then we should plan in advance and give the two different security contexts a distinctive name, for example `containerSecurityContext` and `podSecurityContext` in the CRD.

rubenvp8510:
> > @andreasgerstmayr do you think we need to put a default value?
> 
> We do set a default security context on the containers, for example here:
> 
> https://github.com/grafana/tempo-operator/blob/c39140e6c2b2351ce1baffb88c1cf06a6ad93445/internal/manifests/ingester/ingester.go#L128
> 
> It has some common fields, but is not the same type. I'm not sure if we should set any defaults (`RunAsNonRoot` maybe? but that could break if users use a custom Tempo image).
> 
> Do you think the container security context should also be customizable? Then we should plan in advance and give the two different security contexts a distinctive name, for example `containerSecurityContext` and `podSecurityContext` in the CRD.

I would say we can preserve this without any default value for now. and agree on the renaming.

andreasgerstmayr (approved):

PR #1037, opened 2024-09-25
[Chore] Fix RED metrics test case to ensure monitorTab can display data

IshwarKanse:
The PR fixes the test case to avoid running into issues like https://issues.redhat.com/browse/TRACING-4608 

* The test case was earlier checking only the ClusterRoleBinding with cluster-monitoring-view was attached to the service account used by query frontend. But the test didn't check if the service account is the correct one used by the query frontend deployment. The updated PR ensures that the right service account is created, used by the deployment and has the monitoring view role. 
* Use the tempo-redmetrics-query-frontend SA token to check the metrics from Thanos querier instead of using the prometheus-user-workload-monitoring SA. The query frontend SA has the monitoring view role.

### DIFF
--- a/tests/e2e-openshift/red-metrics/03-assert.yaml
+++ b/tests/e2e-openshift/red-metrics/03-assert.yaml
@@ -23,6 +23,24 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: tempo-redmetrics-query-frontend
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: query-frontend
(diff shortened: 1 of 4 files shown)


IshwarKanse:
```
--- PASS: chainsaw (282.75s)
    --- PASS: chainsaw/red-metrics (282.74s)
PASS
Tests Summary...
- Passed  tests 1
- Failed  tests 0
- Skipped tests 0
Done.
```

iblancasa (approved):

PR #1038, opened 2024-09-25
fix: mount cert and ca to tempo-query

frzifus:
Details: https://issues.redhat.com/browse/TRACING-4703

```bash
$ k get pods -n chainsaw-replicas       
NAME                                           READY   STATUS    RESTARTS   AGE
minio-679d9f746b-hngng                         1/1     Running   0          14m
tempo-cmpreps-compactor-bd74cd5c7-mlq2z        1/1     Running   0          6m2s
tempo-cmpreps-distributor-6bc6b5856f-dv92t     1/1     Running   0          6m2s
tempo-cmpreps-gateway-5897b57875-6wltl         2/2     Running   0          6m2s
tempo-cmpreps-ingester-0                       1/1     Running   0          6m2s
tempo-cmpreps-querier-7b74976fd4-d62hd         1/1     Running   0          6m2s
tempo-cmpreps-query-frontend-5558ffd89-5krzk   3/3     Running   0          6m2s
```

### DIFF
--- a/internal/manifests/queryfrontend/query_frontend.go
+++ b/internal/manifests/queryfrontend/query_frontend.go
@@ -48,11 +48,11 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 
 	if gates.HTTPEncryption || gates.GRPCEncryption {
 		caBundleName := naming.SigningCABundleName(tempo.Name)
-		if err := manifestutils.ConfigureServiceCA(&d.Spec.Template.Spec, caBundleName, 0, 1); err != nil {
+		if err := manifestutils.ConfigureServiceCA(&d.Spec.Template.Spec, caBundleName, 0, 2); err != nil {
 			return nil, err
 		}
 
-		err := manifestutils.ConfigureServicePKI(tempo.Name, manifestutils.QueryFrontendComponentName, &d.Spec.Template.Spec, 0, 1)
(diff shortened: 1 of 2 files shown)


internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -48,11 +48,11 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 
 	if gates.HTTPEncryption || gates.GRPCEncryption {
 		caBundleName := naming.SigningCABundleName(tempo.Name)
-		if err := manifestutils.ConfigureServiceCA(&d.Spec.Template.Spec, caBundleName, 0, 1); err != nil {
+		if err := manifestutils.ConfigureServiceCA(&d.Spec.Template.Spec, caBundleName, 0, 1, 2); err != nil {
```
andreasgerstmayr, line 51: Does the `jaeger-query` container use the certificates?

Not strictly related to this PR, but could we use the container name instead of the index of the container in the deployment?
Because of using the index, the TLS volume mounting broke in the first place :| (because the `jaeger-query` container was added before the `tempo-query` container in the deployment spec).

frzifus: > Does the jaeger-query container use the certificates?

I assume no. But I will check.

> Not strictly related to this PR, but could we use the container name instead of the index of the container in the deployment?
Sure :)

frzifus: @andreasgerstmayr I have added a quick patch on top to determine the container index based on the container names.
In the long term, this functionality should probably be provided in `manifestutils`.

PR #1039, opened 2024-09-26
manifestutils: configure CA and ServicePKI by name

frzifus:
Roughly what you had in  mind?
cc @andreasgerstmayr @IshwarKanse

### DIFF
--- a/internal/manifests/manifestutils/service.go
+++ b/internal/manifests/manifestutils/service.go
@@ -8,6 +8,21 @@ import (
 	"github.com/grafana/tempo-operator/internal/manifests/naming"
 )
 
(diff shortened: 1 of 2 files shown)


andreasgerstmayr:
> Roughly what you had in mind? cc @andreasgerstmayr @IshwarKanse

yes :+1:

Probably we should merge it with https://github.com/grafana/tempo-operator/blob/fce6a2af70615d79c01ac17a94b52fa0bc0d317e/internal/manifests/manifestutils/tls.go at some point, which afaics serves (almost?) the same purpose - mounting TLS certs/keys into the container.

andreasgerstmayr (approved):

PR #1051, opened 2024-10-08
queryfrontend: grant jaeger-query access to pki certs

frzifus:
cc @IshwarKanse

### DIFF
--- a/.chloggen/fix_jaeger-query-certs.yaml
+++ b/.chloggen/fix_jaeger-query-certs.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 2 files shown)


.chloggen/fix_jaeger-query-certs.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack, tempomonolithic
```
andreasgerstmayr, line 5:
```suggestion
component: tempostack
```

only affects tempostack, in tempomonolithic all runs in the same pod and therefore doesn't require mTLS

andreasgerstmayr (approved):

PR #1061, opened 2024-10-15
tempo-query: add ReadinessProbe

### LINKED ISSUE #1058: Add ReadinessProbe for tempo-query
Details:
- https://github.com/grafana/tempo/pull/4178

frzifus:
Closes #1058

Requieres:
- https://github.com/grafana/tempo/pull/4178

### DIFF
--- a/.chloggen/tempo-query_ReadinessProbe.yaml
+++ b/.chloggen/tempo-query_ReadinessProbe.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use the ReadinessProbe to better indicate when tempo-query is ready to accept requests. Improving the startup reliability by avoiding lost data.
+
+# One or more tracking issues related to the change
+issues: [1058]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: Without a readiness check in place, there is a risk that data will be lost when the queryfrontend pod is ready but the tempo query API is not yet available.
--- a/internal/manifests/queryfrontend/query_frontend.go
+++ b/internal/manifests/queryfrontend/query_frontend.go
@@ -311,6 +311,16 @@ func deployment(params manifestutils.Params) (*appsv1.Deployment, error) {
 			},
 			Resources:       tempoQueryResources(tempo),
 			SecurityContext: manifestutils.TempoContainerSecurityContext(),
+			ReadinessProbe: &corev1.Probe{
+				ProbeHandler: corev1.ProbeHandler{
+					GRPC: &corev1.GRPCAction{
+						Port: manifestutils.PortTempoGRPCQuery,
+					},
+				},
+				TimeoutSeconds:   1,
+				PeriodSeconds:    5,
+				FailureThreshold: 12,
+			},
 		}
 		jaegerQueryVolume := corev1.Volume{
 			Name: manifestutils.TmpStorageVolumeName + "-query",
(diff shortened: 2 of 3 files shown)


.chloggen/tempo-query_ReadinessProbe.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
pavolloffay, line 2: is this enhancement or bug_fix?

.chloggen/tempo-query_ReadinessProbe.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add ReadinessProbe to tempo-query
```
pavolloffay, line 8: Why is this being added? What are the consequences of this patch?

.chloggen/tempo-query_ReadinessProbe.yaml
frzifus: From my point of view its a pure enhancement. Everything works as before. We only support the kubernetes readiness check.

.chloggen/tempo-query_ReadinessProbe.yaml
frzifus: Instead of showing that the pod is ready and handling the failure internally, the k8s api will show that its not ready. K8s will do regular calls to the gRPC endpoint to verify that its alive.

frzifus:
> The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.

cc @pavolloffay 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

pavolloffay:
> cc @pavolloffay

> https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

Yes, but does it fixes any particular issues in our deployments?

frzifus:
y, during my multi tenancy debug session I noticed that we sometimes lost some measurements. The test did check for 10 reports but we only had 7-8. Adding a delay solved it. But having the readiness probe made it more resilient.

While typing, I think it would actually be a good Idea to add this kind of check to Jaeger-Query too. There is a extra health endpoint on `localhost:14269/` for it.

Server Response
```bash
{"status":"Server available","upSince":"2024-10-15T08:44:13.514559835Z","uptime":"12.13965331s"}
```

.chloggen/tempo-query_ReadinessProbe.yaml
pavolloffay: Yes, and we should document https://github.com/grafana/tempo-operator/pull/1061#issuecomment-2413275405 in the changelog. Impact of adding readiness probe to our deployment. 

I would also say that this is more a bug_fix than enhancement.

pavolloffay: e.g. improves reliability at startup, avoiding lost data

andreasgerstmayr (approved):

.chloggen/tempo-query_ReadinessProbe.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use the ReadinessProbe to better indicate when tempo-query is ready to accept requests. Improving the startup reliability by avoiding lost data.
+
+# One or more tracking issues related to the change
+issues: [1058]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: Without a readiness check in place, there is a risk that data will be lost when the queryfrontend pod is ready but the tempo query API is not yet available.
```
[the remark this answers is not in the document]
frzifus: The queryfrontend with enabled jaeger contains 3 containers in its pod. tempo-query is part of it.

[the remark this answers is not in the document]
pavolloffay: I understand that, the point I am making is that the 

> data will be lost when the queryfrontend pod

implies to me that this fixes data ingestion, which is not the case

PR #1068, opened 2024-10-22
Bump linter to 1.61.0

rubenvp8510:
I had to bump the linter because old version got stuck in the actions. Not sure what changed, but bumping the version seems to work.

### DIFF
--- a/controllers/tempo/tempostack_controller.go
+++ b/controllers/tempo/tempostack_controller.go
@@ -139,7 +139,7 @@ func (r *TempoStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 //
 //   - For any other error: Set the status condition to Failed,
 //     the Reason to "FailedReconciliation" and the message to the error message.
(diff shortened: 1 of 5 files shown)


iblancasa (approved):

controllers/tempo/tempostack_controller.go
```diff
@@ -139,7 +139,7 @@ func (r *TempoStackReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 //
 //   - For any other error: Set the status condition to Failed,
 //     the Reason to "FailedReconciliation" and the message to the error message.
-func (r *TempoStackReconciler) handleReconcileStatus(ctx context.Context, log logr.Logger, tempo v1alpha1.TempoStack, reconcileError error) (ctrl.Result, error) {
```
rubenvp8510, line 142: @iblancasa @andreasgerstmayr I a comment here to ignore this rule, but might need to change something here?

andreasgerstmayr: I think ignoring the linter rule is fine, it's just complaining that the first return parameter is always `ctrl.Result{}`

andreasgerstmayr (approved):

PR #1069, opened 2024-10-24
Set HTTP protocol in query-frontend ServiceMonitor

rubenvp8510:
When the gateway is disabled, the query frontend is not protected by `mTLS`, nor `TLS`. The only case when it is protected is when the gateway is enabled.

Because of that, the SM needs to be created using HTTP instead of HTTPs, 

Same port/endpoint  is used for Tempo API and the metrics. We cannot choose to protect the `/metrics` endpoint without affect the Tempo API.

cc @andreasgerstmayr  WDYT?

Or perhaps  we should take another approach and try to enable only TLS with the internal certs? 
Or `mTLS` and the `oauth-proxy`? not sure if this scenario is supported by `oauth-proxy.`


andreasgerstmayr (approved):
LGTM :+1:

If we enable TLS using the internal certs we probably break applications which don't load the internal CA bundle. mTLS between oauth-proxy and query-frontend would be nice though (in a different PR), if it's supported.

PR #1081, opened 2024-11-04
Do not add tracing CLI args to opa-openshift container

andreasgerstmayr:
The opa-openshift container does not support
`--internal.tracing.endpoint` and remains in a crashloop if
```
spec:
  observability:
    tracing:
      jaeger_agent_endpoint: ...
      sampling_fraction: "1"
```
is enabled.


pavolloffay:
Is there anything we need to do for the monolithic deployment?

PR #1086, opened 2024-11-07
[Chore] Use OTEL operator with pinned version in upstream CI and fix tests

IshwarKanse:
The PR adds the following changes:

* Use OTEL operator in upstream CI tests using Kind.
* Update tests in tests/e2e testsuite to use the OTEL collector.


IshwarKanse:
The OTEL operator bundle intermittently fails to install in Kind cluster.

IshwarKanse:
@pavolloffay @andreasgerstmayr Can you help to review and approve this PR.

andreasgerstmayr:
> @pavolloffay @andreasgerstmayr Can you help to review and approve this PR.

Let's merge #1092 first (large PR) to not cause merge conflicts

Makefile
```diff
@@ -214,6 +218,14 @@ olm-deploy: operator-sdk ## Deploy operator via OLM
 olm-upgrade: operator-sdk ## Upgrade operator via OLM
 	$(OPERATOR_SDK) run bundle-upgrade -n $(OPERATOR_NAMESPACE) $(BUNDLE_IMG)
 
+##@ OpenTelemetry Operator deployment
+.PHONY: otel-deploy
+otel-deploy: operator-sdk ## Deploy OpenTelemetry operator via OLM
+	kubectl create namespace otel-operator-system
+        # Wait for OLM components to be ready
```
andreasgerstmayr, line 225: Why does the `otel-deploy` target need to wait for the OLM installation, shouldn't the place where OLM is installed wait for completion?

tests/e2e/monolithic-ingestion-mtls/02-install-otel.yaml
```diff
@@ -5,15 +5,32 @@ metadata:
...
+
 ---
-apiVersion: v1
-kind: ConfigMap
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
```
andreasgerstmayr, line 11: The advantage of using a `Deployment` instead of the OpenTelemetry CR is that we don't have a dependency on the OpenTelemetry Operator when running the tests from the e2e test suite. We already have a dependency on the OTEL Operator for the OpenShift e2e test suite.

wdyt @pavolloffay @rubenvp8510, should we mandate an OTEL Operator installation for the Tempo e2e tests? Or keep the dependencies small?
I tend to prefer to keep the dependencies minimal, i.e. just use a `Deployment` for the Tempo e2e tests.

rubenvp8510: I also tend to prefer to keep the dependencies at minimum. What exactly is the advantage of using the OTEL operator instead? It is to not have to bump otel version manually in  each deployment?

IshwarKanse: With a deployment we need to update its spec and related resources if we want to bump the collector version. Its not as simple as changing the image tag ref, for some reason GRPC was not working with deployment when I used the latest collector version and I had to try to figure out what was missing. With operator its much easier, we just update the operator version var in the Makefile (Until the API version changes in CRD :) ) When the tests run on an OpenShift cluster the tests/e2e suite with a deployment runs with the older version of collector. If we use the operator, the tests will run using the operator version installed on the OCP cluster. It could be any version we want to test with Tempo.

rubenvp8510 (approved):

pavolloffay (approved):

PR #1099, opened 2025-01-14
Remove kube-rbac-proxy

rubenvp8510:
Fixes https://github.com/grafana/tempo-operator/issues/1094


.chloggen/remove_metrics_proxy.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
```
andreasgerstmayr, line 2: Does this change affect the end user? Do end users need to update the metric endpoint?
I _think_ it's an internal change and transparent to the user, so I'd mark it as enhancement.

.chloggen/remove_metrics_proxy.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack, tempomonolitic
```
andreasgerstmayr, line 5:
```suggestion
component: operator
```

this change affects the metrics of the operator, not individual operand instances (correct me if I'm wrong)

config/overlays/community/controller_manager_config.yaml
```diff
@@ -4,7 +4,8 @@ distribution: community
 health:
   healthProbeBindAddress: :8081
 metrics:
-  bindAddress: 127.0.0.1:8080
+  secure: true
```
andreasgerstmayr, line 7: If `secure` is `true`, I assume we also need to set `--metrics-tls-cert-file`, is this set somewhere in the community version?

.chloggen/remove_metrics_proxy.yaml
rubenvp8510: It doesn't affect. it should be transparent

config/overlays/community/controller_manager_config.yaml
rubenvp8510: When you don't specific certs those will be self-signed

andreasgerstmayr: ok :+1:

andreasgerstmayr (approved):
I would update the changelog entry from deprecation to enhancement, as for the end user nothing is changing or getting deprecated.

Otherwise, LGTM! Nice change :)

PR #1106, opened 2025-01-30
Set default limits to oauth proxy

rubenvp8510:


internal/manifests/queryfrontend/query_frontend.go
```diff
@@ -90,13 +90,16 @@ func BuildQueryFrontend(params manifestutils.Params) ([]client.Object, error) {
 			jaegerUIAuthentication := tempo.Spec.Template.QueryFrontend.JaegerQuery.Authentication
 
 			if jaegerUIAuthentication != nil && jaegerUIAuthentication.Enabled {
+				defaultOauthProxyResources := manifestutils.Resources(tempo, manifestutils.QueryFrontendComponentName, tempo.Spec.Template.QueryFrontend.Replicas)
```
andreasgerstmayr, line 93: I think it should take the resources from `jaeger-frontend-proxy`, not from `manifestutils.QueryFrontendComponentName`, or am I reading it wrong?

rubenvp8510: Oh true! a mistake! good catch thanks

andreasgerstmayr (approved):

pavolloffay (approved):

PR #1128, opened 2025-02-20
Fix metrics TLS certificate parameters

rubenvp8510:
`CertName` and `KeyName` are only the name of the files, we also need to specify the full path in which those files exists. This PR fix that.

When the certificate file doesn't exist it will self-sign one with _localhost_. Because the path were wrong, it always generate the self signed.

This PR also add a cert-dir flag to specify the path


rubenvp8510:
Upgrade test is failing :/

PR #1131, opened 2025-02-20
Add support for query RBAC to tempomonolithic

pavolloffay:
https://issues.redhat.com/browse/TRACING-5134


api/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -68,9 +68,26 @@ type TempoMonolithicSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Extra Configuration",xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
 	ExtraConfig *ExtraConfigSpec `json:"extraConfig,omitempty"`
 
+	// Query defines query configuration.
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Query Configuration",xDescriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
+	Query MonolithicQuerySpec `json:"query,omitempty"`
```
andreasgerstmayr, line 75:
```suggestion
	Query *MonolithicQuerySpec `json:"query,omitempty"`
```

In the monolithic CR we use pointers for all structs, because otherwise, writing an empty spec to the cluster and reading it back gives two different YAML objects, like this:
```
spec:
  query:
    rbac: {}
```

In the `tempomonolithic_defaults.go` we fill in the possibly-nil pointers to avoid panics when reading from nil pointers: https://github.com/grafana/tempo-operator/blob/78460412bfca629bdd7a84c2317b0182ea8ff92a/api/tempo/v1alpha1/tempomonolithic_defaults.go#L48-L53

api/tempo/v1alpha1/tempomonolithic_types.go
```diff
@@ -68,9 +68,26 @@ type TempoMonolithicSpec struct {
...
 
+// MonolithicQuerySpec defines the query configuration.
+type MonolithicQuerySpec struct {
+	// RBAC defines query RBAC options.
+	// This option can be used only with multi-tenancy.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Query RBAC Settings"
+	RBAC RBACSpec `json:"rbac,omitempty"`
```
andreasgerstmayr, line 88:
```suggestion
	RBAC *RBACSpec `json:"rbac,omitempty"`
```

same as above

PR #1158, opened 2025-04-17
Add cloud cretendial provider S3 support

rubenvp8510:


go.mod
```diff
@@ -26,14 +28,14 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.35.0
 	go.uber.org/zap v1.27.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.29.4
```
andreasgerstmayr, line 29: Could we split the Kubernetes version bump in a separate PR? https://github.com/grafana/tempo-operator/pull/1157 was already created by dependabot. This would make this PR a bit smaller.

Also, we need to do a few things for Kubernetes version bumps: update GitHub actions, add `kind-*.yml` file, and add a new changelog entry.

cmd/root/options.go
```diff
@@ -37,26 +40,32 @@ func loadConfigFile(scheme *runtime.Scheme, outConfig *configv1alpha1.ProjectCon
...
 	err = ctrlConfig.Validate()
 	if err != nil {
-		return nil, options, fmt.Errorf("controller config validation failed: %w", err)
+		return nil, options, manifestutils.TokenCCOAuthConfig{}, fmt.Errorf("controller config validation failed: %w", err)
 	}
 
+	tokenCCOAuth, hasCOOAuthEnv := cloudcredentials.DiscoverTokenCCOAuthConfig()
+
+	// If CCO environment was discovered, override the openshift gateway flag.
+	ctrlConfig.Gates.OpenShift.TokenCCOAuthEnv = hasCOOAuthEnv
```
andreasgerstmayr, line 64: It looks like we don't utilize the feature gate, because it always gets overridden - can we remove the feature gate?
This looks like a good candidate for an `autodiscover` package maybe?

docs/spec/tempo.grafana.com_tempomonolithics.yaml
```diff
@@ -123,6 +123,7 @@ spec:                                    # TempoMonolithicSpec defines the desir
       azure:                             # Azure defines the configuration for Azure Storage.
         secret: ""                       # Secret is the name of a Secret containing credentials for accessing object storage. It needs to be in the same namespace as the TempoMonolithic custom resource.
       backend: "memory"                  # Backend defines the backend for storing traces. Default: memory.
+      credentialMode: ""                 # CredentialMode can be used to set the desired credential mode for authenticating with the object storage. If this is not set, then the operator tries to infer the credential mode from the provided secret and its own configuration.
```
andreasgerstmayr, line 126: Can you move this setting under the `s3` object, as it is only supported for the `S3` backend?

internal/manifests/cloudcredentials/credentialsrequests.go
```diff
@@ -0,0 +1,133 @@
...
+	Controlled     metav1.Object
+	CredentialMode v1alpha1.CredentialMode
+}
+
+// CreateUpdateDeleteCredentialsRequest creates a new CredentialsRequest resource for a TempoStack
+// to request a cloud credentials Secret resource from the OpenShift cloud-credentials-operator.
+func CreateUpdateDeleteCredentialsRequest(ctx context.Context, scheme *runtime.Scheme,
+	params CredentialRequestOptions, k k8client) error {
+
+	if params.CredentialMode != v1alpha1.CredentialModeTokenCCO {
```
andreasgerstmayr, line 40: Is this condition only required for pruning, i.e. switching from CCO back to non-CCO?
When we use `reconcileManagedObjects`, we'd get the pruning functionality for free.

internal/manifests/cloudcredentials/credentialsrequests.go
```diff
@@ -0,0 +1,133 @@
...
+type CredentialRequestOptions struct {
+	TokenCCOAuth   manifestutils.TokenCCOAuthConfig
+	ServiceAccount string
+	Controlled     metav1.Object
+	CredentialMode v1alpha1.CredentialMode
+}
+
+// CreateUpdateDeleteCredentialsRequest creates a new CredentialsRequest resource for a TempoStack
+// to request a cloud credentials Secret resource from the OpenShift cloud-credentials-operator.
+func CreateUpdateDeleteCredentialsRequest(ctx context.Context, scheme *runtime.Scheme,
```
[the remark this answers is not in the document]
rubenvp8510: I tried this, unfortunately I need to make sure to create this object before get the storage parameters . this is because I need the secret that the CCO operator creates  in order to create the rest of the component deployments.

docs/spec/tempo.grafana.com_tempomonolithics.yaml
rubenvp8510: Sure.

cmd/root/root.go
```diff
@@ -27,8 +29,9 @@ type RootConfigKey struct{}
 
 // RootConfig contains configuration relevant for all commands.
 type RootConfig struct {
-	Options    ctrl.Options
-	CtrlConfig configv1alpha1.ProjectConfig
+	Options      ctrl.Options
+	CtrlConfig   configv1alpha1.ProjectConfig
+	TokenCCOAuth *manifestutils.TokenCCOAuthConfig
```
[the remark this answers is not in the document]
rubenvp8510: May be you're right, we don't need to cache things here I will go with the autodiscover package.

internal/manifests/cloudcredentials/credentialsrequests.go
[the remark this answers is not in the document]
rubenvp8510: So the cycle will be Create CCO object ---> getParams -
               -- > if secret still doesn't exist, set the status to degraded (fail as the same as if we don't have storage secret) 
              --> If secret exist, then create the rest of the objects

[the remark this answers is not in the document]
rubenvp8510: Done as we agreed. reusing the reconciliation objects logic.

internal/controller/tempo/tempomonolithic_controller.go
```diff
@@ -37,10 +40,11 @@ import (
...
 	client.Client
-	Scheme     *runtime.Scheme
-	Recorder   record.EventRecorder
-	CtrlConfig configv1alpha1.ProjectConfig
-	Version    version.Version
+	Scheme       *runtime.Scheme
+	Recorder     record.EventRecorder
+	CtrlConfig   configv1alpha1.ProjectConfig
+	Version      version.Version
+	TokenCCOAuth manifestutils.TokenCCOAuthConfig
```
andreasgerstmayr, line 47: Is this field unused?

internal/manifests/manifestutils/params.go
```diff
@@ -15,13 +15,15 @@ type Params struct {
 	TLSProfile          tlsprofile.TLSProfileOptions
 	GatewayTenantSecret []*GatewayTenantOIDCSecret
 	GatewayTenantsData  []*GatewayTenantsData
+	TokenCCOAuth        TokenCCOAuthConfig
```
andreasgerstmayr, line 18: Is this field unused?

internal/controller/tempo/tempostack_controller.go
```diff
@@ -42,10 +44,11 @@ const (
...
 	client.Client
-	Scheme     *runtime.Scheme
-	Recorder   record.EventRecorder
-	CtrlConfig configv1alpha1.ProjectConfig
-	Version    version.Version
+	Scheme       *runtime.Scheme
+	Recorder     record.EventRecorder
+	CtrlConfig   configv1alpha1.ProjectConfig
+	Version      version.Version
+	TokenCCOAuth manifestutils.TokenCCOAuthConfig
```
andreasgerstmayr, line 51: Unused?

andreasgerstmayr (approved):
lgtm
I think the `TokenCCOAuth` field in the controllers and manifests.Params is unused now, as it's stored in the local `tokenCCOAuthEnv := cloudcredentials.DiscoverTokenCCOAuthConfig()` variable?

internal/controller/tempo/tempostack_controller.go
rubenvp8510: oh sorry yes, I forgot to cleanup this.

rubenvp8510:
@IshwarKanse Can we do some QA to this before merge it? It is a little bit big PR.

IshwarKanse:
@rubenvp8510 Yes, sure. I'll have the results by tomorrow EOD as I have few tasks today.

IshwarKanse:
@rubenvp8510 
I am getting the following error in Tempo CR
cannot configure tempo in CCO mode without CCO environment
when I set credentialMode: token-cco
This is the draft test PR: 
https://github.com/grafana/tempo-operator/pull/1189/files#diff-de78868cf0ad26770fe7daccc986c5e26ec395da6b4746acb5ba553ca5bfb30cR12

IshwarKanse:
@rubenvp8510 After adding the ROLEARN to operator subscription, the CCO is working for TempoStack. However for monolithic, the monolithic pod fails with following error. This is happening for both STS and STS with CCO. 

  Warning  Failed                  2m2s (x8 over 2m58s)  kubelet                  Error: couldn't find key access_key_secret in Secret

Also, for the TempoStack with CCO, the service accounts created for the query frontend instance has the annotations set.
    eks.amazonaws.com/audience: sts.amazonaws.com
    eks.amazonaws.com/role-arn: 
 This doesn't affect the functionality but we should remove it when running with CCO, we don't have it in the other SA for Tempo instance. There are two SA created one for Tempo instance and the other for query frontend.

rubenvp8510:
@IshwarKanse thanks for testing, this should be ready with the fixes for the issues you found.

IshwarKanse:
@rubenvp8510 Thanks, I'll retest and run some regression tests today.

PR #1169, opened 2025-04-29
Harden CI: resolve zizmor alerts

mapno:
changed based on audit using zizmor: https://woodruffw.github.io/zizmor/audits/

see more: https://grafana.com/blog/2025/04/27/grafana-security-update-no-customer-impact-from-github-workflow-vulnerability/

### DIFF
--- a/.github/workflows/publish-test-utils-image.yaml
+++ b/.github/workflows/publish-test-utils-image.yaml
@@ -56,5 +56,3 @@ jobs:
           platforms: linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
           push: ${{ github.event_name == 'push' }}
           tags: ${{ steps.meta.outputs.tags }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache
--- a/.github/workflows/publish-test-utils-image.yaml
+++ b/.github/workflows/publish-test-utils-image.yaml
@@ -17,6 +17,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   test-utils:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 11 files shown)


javiermolinar (approved):

.github/workflows/publish-test-utils-image.yaml
```diff
@@ -17,43 +17,39 @@ concurrency:
...
           images: ghcr.io/${{ github.repository }}/test-utils
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-
-      - name: Cache Docker layers
```
andreasgerstmayr, line 40: Why was this step removed, is this caching step a security issue?

.github/workflows/reusable-publish-images.yaml
```diff
@@ -38,28 +43,20 @@ jobs:
             type=ref,event=branch
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-
-      - name: Cache Docker layers
```
andreasgerstmayr, line 46: Why was this step removed, is this caching step a security issue?

.github/workflows/publish-test-utils-image.yaml
andreasgerstmayr: Maybe because it's in the publish GitHub action, in that case, we should also remove `cache-from` and `cache-to` from the `docker/build-push-action` action.

.github/workflows/reusable-publish-images.yaml
electron0zero: these steps were flagged as https://woodruffw.github.io/zizmor/audits/#cache-poisoning issue.

andreasgerstmayr: Ok, I've no objections to removing the cache step in the publish workflow.

Could you also remove the references to `/tmp/.buildx-cache` in the same workflow? (`cache-from` and `cache-to` of the `docker/build-push-action` action, and a workaround to move the cache)

.github/workflows/publish-test-utils-image.yaml
mapno: Yes, as Suraj pointed out the cache action is vulnerable to [cache poisoning](https://woodruffw.github.io/zizmor/audits/#cache-poisoning). I figured it wasn't executed that frequently that justified keeping it.

>we should also remove cache-from and cache-to from the docker/build-push-action action.

Yes, nice catch.

andreasgerstmayr (approved):

rubenvp8510 (approved):

PR #1185, opened 2025-05-07
RBAC: allow cluster admins to see all attributes

pavolloffay:
Resolves https://issues.redhat.com/browse/TRACING-5354

### DIFF
--- a/.chloggen/fix-cluster-admin-filtering.yaml
+++ b/.chloggen/fix-cluster-admin-filtering.yaml
@@ -1,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack, tempomonolithic
+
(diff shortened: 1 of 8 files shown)


pavolloffay:
@IshwarKanse could you please test this PR on OCP? I am not sure if there are any consequences. We should also test the Jaeger UI with if the auth works well (kubeadmin can by default see everything).

IshwarKanse:
@pavolloffay Tested and we can merge this PR.

Here are the tests performed:
TempoStack and TempoMonolithic without RBAC:
* Added traces reader role to testuser-0
* testuser-0 is able to view complete traces from Jaeger UI and Tracing UI.
* kubeadmin is able to view complete traces from Jager UI and Tracing UI.
* testuser-1 is denied access to the tenant as expected.

TempoStack and TempoMonolithic with RBAC enabled:
* kubeadmin can view complete traces, from Tracing UI.
* testuser-0 can view complete traces from its namespace, from Tracing UI.
* testuser-1 cannot view any traces since it doesn't have access to any namespace which is generating traces. 
* testuser-2 with cluster-admin cluster role can view complete traces, from Tracing UI.

andreasgerstmayr (approved):

PR #1195, opened 2025-05-16
[chore] add permissions on job level

andreasgerstmayr:
Both the "root" workflow and the reusable workflow have to have the permissions set. Additionally, I moved the permissions on the job level instead of the root level (which would apply to all jobs).


andreasgerstmayr:
I tried this in my fork and it finally works

rubenvp8510 (approved):

PR #1197, opened 2025-05-17
Set GOMEMLIMIT to 80% of memory limit, if any

rubenvp8510:


internal/manifests/manifestutils/gomemlimit.go
```diff
@@ -0,0 +1,28 @@
+package manifestutils
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// SettGoMemLimit sets GOMEMLIMIT env var to 80% memory of the container if it's defined.
+func SettGoMemLimit(containerName string, pod *v1.PodSpec) {
```
andreasgerstmayr, line 10:
```suggestion
func SetGoMemLimit(containerName string, pod *v1.PodSpec) {
```

typo

andreasgerstmayr (approved):

andreasgerstmayr:
Could you update an unit test and an e2e test with the new env var?

rubenvp8510:
@andreasgerstmayr Do you think we need to add it to the e2e? or is enough with the unit tests?

Issue #1120, opened 2025-02-14
Cannot create metrics-generator because of read-only filesystem


Jeansen:
I have the following part in my a monolith operator yaml:

```
...
  extraConfig:                           
    tempo:
      metrics_generator:
        processor:
          span_metrics:
            enable_target_info: true
          local_blocks:
            filter_server_spans: false
            flush_to_storage: true
        storage:
          path: "/tmp"
          remote_write:
            - url: http://grafana-mimir-gateway.default.svc/api/v1/push
              send_exemplars: true
...
```

In the logs, I see that everything comes up fine, except:

```
level=warn ts=2025-02-14T23:42:15.753098235Z caller=server.go:1187 method=/tempopb.MetricsGenerator/PushSpans duration=211.297µs msg=gRPC err="could not create directory for metrics WAL: mkdir /tmp/single-tenant: read-only file system"
```

The operator gives me no way to add additional volumes or overwrite the read-only root file system setting. So, currently there is no way to activate the metrics_generator because it needs a path for the WAL to write to.

andreasgerstmayr:
Can you update your config to use `/var/tempo/generator/wal` as the path for the WAL? `/var/tempo` is mounted as a tmpfs for in-memory-mode, and as a PV otherwise.

In the past, I used the following spec successfully:
```
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoMonolithic
metadata:
  name: sample
spec:
  extraConfig:
    tempo:
      metrics_generator:
        storage:
          path: /var/tempo/generator/wal
          remote_write:
          - url: http://kube-prometheus-stack1-prometheus.kube-prometheus-stack1:9090/api/v1/write
            send_exemplars: true
      overrides:
        defaults:
          metrics_generator:
            processors: [service-graphs, span-metrics]
```

dobesv:
Were you able to get metrics-generator working with the tempo-operator?  I'm trying to figure this out.  I guess this might work with the "monolithic" setup because it's running everything in one process instead of telling each pod to run just one component, so tempo will run the metrics-generator within that monolithic deployment?

Starttoaster:
Is there any plan to support metrics-generator in a TempoStack deployment?

PR #1401, opened 2026-03-06
feat: add support for metrics-generator

nickcaballero:
Add support for deploying metrics-generator via TempoStack resource. The original operator [design doc](https://docs.google.com/document/d/1avSSf__R226l2b3hbcpXlYH7w6iKtXZsd9VTcpxDqng/edit?pli=1&tab=t.0#heading=h.cmtg3tpfu85t) doesn't mention why the metrics-generator wasn't included, but it is required to support some queries.

Related:
- https://github.com/grafana/tempo-operator/issues/1120


nickcaballero:
Working on fixing checks.

nickcaballero:
@andreasgerstmayr Fixed lint issue, added changelog entry, and updated bundle output.

jkranner:
Hi Tempo team,
I would really like to use the metrics-generator as part of the TempoStack setup with the Tempo Operator.
While looking into this, I noticed that the version currently published on **operatorhub.io** seems to be quite a bit **older** and **doesn’t include support for the metrics-generator** yet. **Would it be possible to publish an updated operator version there as well, once this functionality is fully ready?**
That would make it much easier for users who rely on OperatorHub/OLM to access newer Tempo features.
Thanks a lot for your great work on Tempo, and for taking a look at this!
Best regards

internal/manifests/config/build.go
```diff
@@ -133,6 +133,19 @@ func buildConfiguration(params manifestutils.Params) ([]byte, error) {
 		opts.TenantRateLimitsPath = tenantOverridesMountPath
 	}
 
+	if tempo.Spec.Template.MetricsGenerator != nil {
+		processors := tempo.Spec.Template.MetricsGenerator.Processors
+		if len(processors) == 0 {
+			processors = []string{"service-graphs", "span-metrics"}
```
andreasgerstmayr: Can you move setting the defaults to `tempostack_webhook.go`?

internal/manifests/config/tempo-config.yaml
```diff
@@ -115,6 +115,23 @@ distributor:
...
+      tls_key_path: {{ .TLS.Paths.Key }}
+      tls_ca_path: {{ .TLS.Paths.CA }}
+      tls_server_name: {{ .MetricsGenerator.Address }}
+{{- if .TLS.Profile.Ciphers }}
+      tls_cipher_suites: {{ .TLS.Profile.Ciphers }}
+{{- end }}
+      tls_min_version: {{ .TLS.Profile.MinTLSVersion }}
+{{- end }}
+    address: {{ .MetricsGenerator.Address }}
+  generate_all_spans: true
```
andreasgerstmayr: Where is `generate_all_spans` documented? https://grafana.com/docs/tempo/v2.10.x/configuration/manifest/

internal/manifests/config/tempo-config.yaml
```diff
@@ -115,6 +115,23 @@ distributor:
     {{- with .MemberList.InstanceAddr }}
     instance_addr: {{ . }}
     {{- end }}
+{{- if .MetricsGenerator.Enabled }}
+  metric_generator_client:
```
andreasgerstmayr: Afaics this is a top-level setting and called `metrics_generator_client`: https://grafana.com/docs/tempo/v2.10.x/configuration/manifest/

.chloggen/feat_metrics-generator.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for deploying metrics-generator via TempoStack resource.
+
+# One or more tracking issues related to the change
+issues: [1120]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
andreasgerstmayr: Could you add a small example `TempoStack` spec how to enable it?

internal/manifests/metricsgenerator/metricsgenerator.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/grafana/tempo-operator/internal/manifests/manifestutils"
+	"github.com/grafana/tempo-operator/internal/manifests/memberlist"
+	"github.com/grafana/tempo-operator/internal/manifests/naming"
+)
+
+// BuildMetricsGenerator creates metrics-generator objects.
+func BuildMetricsGenerator(params manifestutils.Params) ([]client.Object, error) {
+	d := deployment(params)
+
+	err := error(nil)
```
andreasgerstmayr:
```suggestion
	var err error
```

andreasgerstmayr (commented):
Hi! Sorry for the late review.

Thank you for the PR, it looks good overall. I just had a few small follow-ups:
- please add a call to `manifestutils.PatchEnvVars()`, similar to the other components
- update the generated network policies
- add basic unit tests for the `metricsgenerator` package

internal/manifests/config/tempo-config.yaml
nickcaballero: It's not, hallucinated 🤡

andreasgerstmayr:
I just tested this PR in a Kubernetes cluster with the following CR:
```
kind: TempoStack
metadata:
  name: simplest
spec:
  storage:
    secret:
      name: minio-test
      type: s3
  storageSize: 1Gi
  resources:
    total:
      limits:
        memory: 2Gi
        cpu: 2000m
  template:
    queryFrontend:
      jaegerQuery:
        enabled: true
    metricsGenerator:
      remoteWriteURLs:
        - http://prometheus:9090/api/v1/write
```

but all Tempo pods are in a crashloop:
```
failed parsing config: failed to parse configFile /conf/tempo.yaml: yaml: unmarshal errors:
  line 121: field address not found in type client.Config
  line 127: field service-graphs not found in type generator.ProcessorConfig
  line 128: field span-metrics not found in type generator.ProcessorConfig
```

Can you share how you tested this change?

nickcaballero:
> Can you share how you tested this change?

Apologies, @andreasgerstmayr. I was relying heavily on unit tests and missed a few things. After testing locally via k3s, I've updated the PR with the following changes:

- Removed configuration of processors from CRD and updated example to use `extraConfig.tempo` instead
- Fixed missing mTLS secret
- Added storage access to metrics-generator
- Added network policy

sammykrystal:
We'd also really like to make use of this metrics-generator being in the tempo-operator as it doesn't seem easy to deploy it separately without moving away from the operator approach. Appreciate the work in progress here.

andreasgerstmayr (changes requested):
Thanks for the changes. I tested it again, the pods are now running, but no metrics are generated.

The problem is that metrics-generator is deployed, but not enabled. Take a look at https://github.com/grafana/tempo/blob/v2.10.5/example/docker-compose/shared/tempo.yaml on how to enable it.

For reference, I'm using the following manifest to deploy TempoStack with metrics-generator, Prometheus and Grafana:
```
apiVersion: v1
kind: Secret
metadata:
  name: minio-test
type: Opaque
stringData:
  endpoint: http://minio.minio.svc:9000
  bucket: tempo
  access_key_id: tempo
  access_key_secret: supersecret
---
kind: TempoStack
metadata:
  name: simplest
spec:
  storage:
    secret:
      name: minio-test
      type: s3
  storageSize: 1Gi
  template:
    metricsGenerator:
      remoteWriteURLs:
      - http://prometheus:9090/api/v1/write
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: prometheus
  labels:
    app.kubernetes.io/name: prometheus
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: prometheus
  template:
    metadata:
      labels:
        app.kubernetes.io/name: prometheus
    spec:
      containers:
      - name: prometheus
        image: prom/prometheus:latest
        args:
        - --config.file=/etc/prometheus/prometheus.yml
        - --web.enable-remote-write-receiver
        volumeMounts:
        - name: config
          mountPath: /etc/prometheus
      volumes:
      - name: config
        configMap:
          name: prometheus-config
---
apiVersion: v1
kind: Service
metadata:
  name: prometheus
spec:
  selector:
    app.kubernetes.io/name: prometheus
  ports:
  - port: 9090
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: prometheus-config
data:
  prometheus.yml: |
    global:
      scrape_interval: 15s
    scrape_configs:
      - job_name: prometheus
        static_configs:
          - targets: ["localhost:9090"]
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: grafana
  labels:
    app.kubernetes.io/name: grafana
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: grafana
  template:
    metadata:
      labels:
        app.kubernetes.io/name: grafana
    spec:
      containers:
      - name: grafana
        image: grafana/grafana:latest
        env:
        - name: GF_AUTH_ANONYMOUS_ENABLED
          value: "true"
        - name: GF_AUTH_ANONYMOUS_ORG_ROLE
          value: Admin
        volumeMounts:
        - name: datasources
          mountPath: /etc/grafana/provisioning/datasources
      volumes:
      - name: datasources
        configMap:
          name: grafana-datasources
---
apiVersion: v1
kind: Service
metadata:
  name: grafana
spec:
  selector:
    app.kubernetes.io/name: grafana
  ports:
  - port: 3000
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: grafana-datasources
data:
  datasources.yaml: |
    apiVersion: 1
    datasources:
      - name: Prometheus
        type: prometheus
        uid: prometheus
        url: http://prometheus:9090
        isDefault: true
      - name: Tempo
        type: tempo
        uid: tempo
        url: http://tempo-simplest-query-frontend:3200
        jsonData:
          serviceMap:
            datasourceUid: prometheus
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: k6-tracing
  labels:
    app.kubernetes.io/name: k6-tracing
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: k6-tracing
  template:
    metadata:
      labels:
        app.kubernetes.io/name: k6-tracing
    spec:
      containers:
      - name: k6-tracing
        image: ghcr.io/grafana/xk6-client-tracing:v0.0.5
        env:
        - name: ENDPOINT
          value: tempo-simplest-distributor:4317
```

nickcaballero:
Hey @andreasgerstmayr! So I opted to remove the processor configuration from the CRD. Instead, I updated the example in the change notes to use `extraConfig.tempo` instead. My reasoning was that there are many options that can be configured for the processors, all of which would simply be copied into the configuration file, just as `extraConfig.tempo` already allows.

andreasgerstmayr:
> Hey @andreasgerstmayr! So I opted to remove the processor configuration from the CRD. Instead, I updated the example in the change notes to use `extraConfig.tempo` instead. My reasoning was that there are many options that can be configured for the processors, all of which would simply be copied into the configuration file, just as `extraConfig.tempo` already allows.

I see your point, and agree that there are too many options which we won't copy straight to the CRD. We added `extraConfig` as an escape hatch for overwriting the generated Tempo configuration with any of the hundreds of configuration options.

However, it is not meant for the configuration of supported features. In other words, the basic, supported functionality of the operator should be configurable in the CRD (in this case, the `metricsGenerator` property). The UX would suffer if we ask users to configure parts in the CRD, and other parts of the same feature in the `extraConfig`.

Please add back the configuration of the processors, i.e. the basic functionality to make it work and generate metrics. Once that is done, the PR is good to merge :)

nickcaballero:
@andreasgerstmayr I was reviewing the documentation, and it looks like the overrides property changed after Tempo 2.3: `overrides.metrics_generator_processors` changes to `overrides.defaults.metrics_generator.processors`. It looks like the configuration currently being generated is for Tempo < 2.3, but the Tempo image being used is 2.10. Is that because Tempo versions >= 2.3 still support the old overrides configuration?

andreasgerstmayr:
> It looks like the configuration currently being generated is for Tempo < 2.3, but the Tempo image being used is 2.10. Is that because Tempo versions >= 2.3 still support the old overrides configuration?

I think we forgot to update to the new configuration. Yep, seems like Tempo still supports the old overrides, otherwise it would crash at startup afaics.

cc @rubenvp8510 we should update the overrides config (see https://grafana.com/docs/tempo/latest/release-notes/v2-3/#changes-to-the-overrides-module-configuration)

nickcaballero:
Sounds good. I'll follow the legacy syntax for the overrides, since it sounds like there might be a wider effort to migrate overrides between versions.

nickcaballero:
@andreasgerstmayr This is ready to test again. I've updated the changelog to include a sample configuration that worked for me during local testing.

andreasgerstmayr (dismissed):
Thanks for the updates! I added a few changes to conform with the rest of the operator:
- use `.enabled` field (similar to enabling the gateway, jaeger query etc.)
- set storage paths
- utilize default and validating webhooks
- remove metrics-generator deployment and service if `.metricsGenerator.enabled` is updated from `true` to `false`
- add example file to `docs/examples`

I tested it on both Kubernetes and OpenShift, and it's working well :)
Thanks for your contribution!

andreasgerstmayr (approved):

rubenvp8510 (approved):

PR #1415, opened 2026-03-30
Fix OTLP HTTP receiver to use internal mTLS when gateway and HTTPEncryption are enabled 

IshwarKanse:
The PR fixes issue: https://redhat.atlassian.net/browse/TRACING-6133

When gateway and HTTPEncryption were both enabled, the OTLP HTTP receiver was configured with the gateway serving certificate (which only has SANs for the gateway service DNS names). Since the gateway connects to the receiver at localhost:4318, TLS verification failed with "certificate is valid for tempo-*-gateway.*.svc, not localhost", causing all OTLP HTTP trace ingestion to be silently dropped.
  
The gRPC receiver already had a special case for this scenario, using the internal mTLS certificate (which includes localhost in its SANs). This commit adds the same handling for the HTTP receiver.

### DIFF
--- a/.chloggen/fix_monolithic_otlp_http_receiver_tls.yaml
+++ b/.chloggen/fix_monolithic_otlp_http_receiver_tls.yaml
@@ -1,0 +1,23 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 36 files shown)


IshwarKanse:
cc @rubenvp8510 @andreasgerstmayr

rubenvp8510 (approved):

IshwarKanse:
Also updated the tests.

pavolloffay (approved):

PR #1423, opened 2026-04-13
Fix race condition causing TempoMonolithic stuck in terminating state

IshwarKanse:
## Summary
- Fix a race condition where cert-hash annotation updates during deletion cause TempoMonolithic (and TempoStack) resources to get stuck in terminating state for 600+ seconds
- Add `DeletionTimestamp` guards to the cert rotation controllers to prevent annotation updates on resources being deleted
- Add unit tests for both cert rotation controllers
- Reorder TLS profile e2e tests so override tests run before the APIServer-level test

## Root Cause
The tempo-operator has separate cert rotation controllers that update CR annotations during certificate operations:

1. **`CertRotationReconciler`** — watches TempoStack CRs for expired certificates and updates `tempo.grafana.com/certRotationRequiredAt` annotation
2. **`CertRotationMonolithicReconciler`** — same for TempoMonolithic CRs

Neither checked `DeletionTimestamp` before updating the CR. When a resource is deleted:
1. Kubernetes sets `DeletionTimestamp` and adds `foregroundDeletion` finalizer
2. The cert rotation controllers still run and update annotations on the CR
3. This changes the `resourceVersion`, causing conflicts with the finalizer removal
4. The resource gets stuck in terminating state indefinitely

Note: The main `TempoStackReconciler` and `TempoMonolithicReconciler` controllers already had `DeletionTimestamp` guards, so the functions they call (`CreateOrRotateCertificates`, etc.) were already protected. The bug was only in the independent cert rotation controllers.

## Changes
| File | Change |
|------|--------|
| `internal/controller/tempo/certrotation_controller.go` | Add `DeletionTimestamp` check before cert rotation reconciliation for TempoStack |
| `internal/controller/tempo/certrotation_monolithic_controller.go` | Add `DeletionTimestamp` check before cert rotation reconciliation for TempoMonolithic |
| `internal/controller/tempo/certrotation_controller_test.go` | Add unit tests for both controllers: deleting, not-found, and unmanaged scenarios |
| `tests/e2e-openshift-tls-profile/` | Rename test dirs with numeric prefixes to control execution order (override tests first, APIServer test last) |

## Impact
- **TempoMonolithic**: No longer gets stuck in terminating state due to cert rotation annotation updates
- **TempoStack**: Same fix applied proactively
- **Downstream**: opendatahub-operator had to implement a workaround (force finalizer removal) — see opendatahub-io/opendatahub-operator#3351


internal/manifests/monolithic/certificates.go
```diff
@@ -39,6 +39,13 @@ func CreateOrRotateCertificates(ctx context.Context, log logr.Logger,
 		return kverrors.Wrap(err, "failed to lookup tempomonolithic", "name", req.String())
 	}
 
+	// Skip certificate operations if the resource is being deleted to avoid
+	// conflicts with the foregroundDeletion finalizer.
+	if stack.GetDeletionTimestamp() != nil {
```
andreasgerstmayr: This `GetDeletionTimestamp()` is already checked in https://github.com/grafana/tempo-operator/blob/4913e7e3c4970733e0f44d46204e25c842283ad8/internal/controller/tempo/tempomonolithic_controller.go#L81

internal/certrotation/handlers/rotate_certs.go
```diff
@@ -38,6 +38,13 @@ func CreateOrRotateCertificates(ctx context.Context, log logr.Logger,
 		return kverrors.Wrap(err, "failed to lookup tempostacks", "name", req.String())
 	}
 
+	// Skip certificate operations if the resource is being deleted to avoid
+	// conflicts with the foregroundDeletion finalizer.
+	if stack.GetDeletionTimestamp() != nil {
```
andreasgerstmayr: `GetDeletionTimestamp()` is already checked in `tempostack_controller.go`

internal/certrotation/handlers/certrotation_discovery.go
```diff
@@ -30,6 +30,12 @@ func AnnotateTempoStackForRequiredCertRotation(ctx context.Context, k client.Cli
 		return kverrors.Wrap(err, "failed to get tempo TempoStack", "key", key)
 	}
 
+	// Skip annotation update if the resource is being deleted to avoid
+	// conflicts with the foregroundDeletion finalizer.
+	if s.GetDeletionTimestamp() != nil {
```
andreasgerstmayr: already checked in `certrotation_controller.go` (in your PR)

internal/certrotation/handlers/certrotation_discovery.go
```diff
@@ -60,6 +66,12 @@ func AnnotateMonolithicForRequiredCertRotation(ctx context.Context, k client.Cli
...
```
andreasgerstmayr: already checked in `certrotation_monolithic_controller.go` (in your PR)

andreasgerstmayr (commented):
Let's keep the new `GetDeletionTimestamp()` checks in the `_controller.go` files, but drop the others (these are dead code, as the functions are called from the controller, which has the identical check).

internal/manifests/monolithic/certificates.go
IshwarKanse: Agreed, removed. This function has exactly one caller (the controller), which already guards against deletion. Also added unit tests for both cert rotation controllers to verify they skip reconciliation for deleting resources.

internal/certrotation/handlers/rotate_certs.go
IshwarKanse: Agreed, removed. The sole caller (`tempostack_controller.go`) already checks `DeletionTimestamp` before reaching this function.

internal/certrotation/handlers/certrotation_discovery.go
IshwarKanse: Agreed, removed. `certrotation_controller.go` is the only caller and already returns early when `DeletionTimestamp` is set.

internal/certrotation/handlers/certrotation_discovery.go
IshwarKanse: Agreed, removed. `certrotation_monolithic_controller.go` is the only caller and already returns early when `DeletionTimestamp` is set.

andreasgerstmayr (approved):

PR #1440, opened 2026-05-08
Fix race condition in certificate hash annotation update

pavolloffay:
## Problem

The `e2e/compatibility` test was flaky due to a race condition in `addCertificateHashAnnotations()`. The function updated the TempoStack object using a stale copy fetched at the beginning of the reconciliation loop. If the TempoStack was modified between the initial `Get` and the annotation `Update` (e.g. by a concurrent status update), the update failed with a resource version conflict:

```
built in cert manager error: failed to add certificate hash annotations:
Operation cannot be fulfilled on tempostacks.tempo.grafana.com "simplest":
the object has been modified; please apply your changes to the latest version and try again
```

This conflict error was not retried — it propagated up to `handleReconcileStatus`, which treated it as a generic error and set a `FailedReconciliation` condition on the TempoStack. Although the controller would eventually requeue and succeed, the transient `Failed` condition caused assertion timing issues in e2e tests.

## Fix

- Wrap the TempoStack annotation update in `retry.RetryOnConflict` with a fresh `Get` before each retry attempt, so resource version conflicts are handled gracefully instead of surfacing as reconciliation failures.
- Increase the assert timeout for the `tshirt-size-pico` e2e test to 5 minutes to account for slow CI environments.

### DIFF
--- a/internal/certrotation/handlers/rotate_certs.go
+++ b/internal/certrotation/handlers/rotate_certs.go
@@ -11,6 +11,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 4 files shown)


andreasgerstmayr (dismissed):

frzifus (dismissed):

internal/certrotation/handlers/rotate_certs.go
```diff
@@ -116,6 +118,14 @@ func addCertificateHashAnnotations(ctx context.Context, k client.Client, stack *
 		certSecrets[name] = cert.Secret
 	}
 
-	stack.Annotations = labels.Merge(stack.Annotations, manifestutils.CertificateHashAnnotations(certSecrets))
-	return k.Update(ctx, stack)
+	hashAnnotations := manifestutils.CertificateHashAnnotations(certSecrets)
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
andreasgerstmayr, line 123: if this resolves the issue, can you do the same fix in `monolithic/certificates.go`?

pavolloffay: yes, I will restart the jobs a couple of times

pavolloffay: 5 runs without failure. it seems to work.

I have applied the fix for the monolithic as well

andreasgerstmayr (approved):

PR #1453, opened 2026-05-22
[chore] add CODEOWNERS for config paths

electron0zero:
Require reviews from core maintainers for changes to workflows, Makefile, Dockerfile, go.mod, and go.sum.


mapno (dismissed):

mapno (commented):
Should it include CODEOWNERS itself? So permissions can't be changed without maintainers approval either.

electron0zero:
> Should it include CODEOWNERS itself? So permissions can't be changed without maintainers approval either.

@mapno good catch, added it.

andreasgerstmayr (approved):

PR #1491, opened 2026-06-01
fix: mark managementState and tenants.mode optional

timgriffiths:
## Description

`TempoStackSpec.ManagementState` and `TenantsSpec.Mode` each declare a `+kubebuilder:default` (`Managed` and `static` respectively) **and** are marked `+required` / `+kubebuilder:validation:Required`. A field that the API server always populates via structural-schema defaulting cannot meaningfully be a required *input*.

The practical effect is that static schema validators that don't apply defaults — e.g. `kubeconform`, used widely in GitOps/CI pipelines — reject otherwise-valid resources that omit these fields:

```
TempoStack foo is invalid: jsonschema validation failed
  - at '/spec': missing property 'managementState'
  - at '/spec/tenants': missing property 'mode'
```

These resources apply cleanly and run correctly against a live cluster (the API server defaults the fields before validation) — only offline validators trip on them.

## Changes

- Mark `ManagementState` and `Mode` `+optional` / `+kubebuilder:validation:Optional`, keeping their `+kubebuilder:default` markers.
- Add `omitempty` to the `Mode` json tag for consistency with other optional+defaulted fields (e.g. `ManagementState`).
- Regenerate CRDs and OLM bundle manifests.

Behaviour is unchanged: the API server still defaults both fields to `Managed` / `static`. The only difference is the generated CRD no longer lists them under `required`, so `TempoStack` and `TempoMonolithic` (which shares `TenantsSpec`) validate cleanly with offline tooling.

The diff is limited strictly to the two markers and the `required` lists they drive in the generated manifests.

### DIFF
--- a/.chloggen/optional-defaulted-required-fields.yaml
+++ b/.chloggen/optional-defaulted-required-fields.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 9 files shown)


andreasgerstmayr (approved):

andreasgerstmayr:
Thanks for your contribution!

PR #1515, opened 2026-06-15
[chore] fix: harden must-gather operator discovery

immanuwell:
this fixes 2 small must-gather footguns.

`--operator-name` was ignored, so the code always looked for `app.kubernetes.io/name=tempo-operator`.
also, if the operator Deployment existed but had no matching Pods yet, `must-gather` indexed `operatorPods.Items[0]` and blew up.

repro
1. run `must-gather` while the operator Deployment exists but the pod is temporarily gone. pretty easy to hit during rollout, `ImagePullBackOff`, `CrashLoopBackOff`, `Pending`, or right after deleting the pod.
2. before this patch, it panics on `operatorPods.Items[0]`.
3. run `./bin/must-gather --operator-name foo`.
4. before this patch, it still searched for `tempo-operator`.

now it honors `--operator-name` and `--operator-namespace`, and returns a clear error when no operator pod is found.

### DIFF
--- a/cmd/gather/cluster/cluster.go
+++ b/cmd/gather/cluster/cluster.go
@@ -60,19 +60,32 @@ func (c *cluster) getOperatorNamespace() (string, error) {
 }
 
 func (c *cluster) getOperatorDeployment() (appsv1.Deployment, error) {
-	operatorDeployments := appsv1.DeploymentList{}
-	err := c.config.KubernetesClient.List(context.TODO(), &operatorDeployments, &client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(labels.Set{
-			"app.kubernetes.io/name": "tempo-operator",
+	listOptions := []client.ListOption{
+		client.MatchingLabels(map[string]string{
+			"app.kubernetes.io/name": c.config.OperatorName,
 		}),
-	})
+	}
+	if c.config.OperatorNamespace != "" {
+		listOptions = append(listOptions, client.InNamespace(c.config.OperatorNamespace))
+	}
+
+	operatorDeployments := appsv1.DeploymentList{}
+	err := c.config.KubernetesClient.List(context.TODO(), &operatorDeployments, listOptions...)
 
 	if err != nil {
 		return appsv1.Deployment{}, err
 	}
 
 	if len(operatorDeployments.Items) == 0 {
-		return appsv1.Deployment{}, fmt.Errorf("operator not found")
+		return appsv1.Deployment{}, fmt.Errorf("operator deployment with app.kubernetes.io/name=%q not found", c.config.OperatorName)
+	}
+
+	if len(operatorDeployments.Items) > 1 {
+		deployments := make([]string, 0, len(operatorDeployments.Items))
+		for _, deployment := range operatorDeployments.Items {
+			deployments = append(deployments, fmt.Sprintf("%s/%s", deployment.Namespace, deployment.Name))
+		}
+		return appsv1.Deployment{}, fmt.Errorf("found multiple operator deployments for app.kubernetes.io/name=%q: %s", c.config.OperatorName, strings.Join(deployments, ", "))
 	}
 
 	return operatorDeployments.Items[0], nil
@@ -93,6 +106,10 @@ func (c *cluster) GetOperatorLogs() error {
 	})
 	if err != nil {
 		return err
+	}
+
+	if len(operatorPods.Items) == 0 {
+		return fmt.Errorf("no pods found for operator deployment %s/%s", deployment.Namespace, deployment.Name)
 	}
 
 	pod := operatorPods.Items[0]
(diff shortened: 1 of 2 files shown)


cmd/gather/cluster/cluster.go
```diff
@@ -60,19 +60,32 @@ func (c *cluster) getOperatorNamespace() (string, error) {
 }
 
 func (c *cluster) getOperatorDeployment() (appsv1.Deployment, error) {
-	operatorDeployments := appsv1.DeploymentList{}
-	err := c.config.KubernetesClient.List(context.TODO(), &operatorDeployments, &client.ListOptions{
-		LabelSelector: labels.SelectorFromSet(labels.Set{
-			"app.kubernetes.io/name": "tempo-operator",
+	listOptions := []client.ListOption{
+		client.MatchingLabels(map[string]string{
+			"app.kubernetes.io/name": c.config.OperatorName,
```
andreasgerstmayr, line 65: This label is hardcoded in the operator and won't change: https://github.com/grafana/tempo-operator/blob/574e2e864e1f617e9eb36a9bb5cb6ea0c6fcd590/internal/manifests/manifestutils/labels.go#L42

andreasgerstmayr (approved):
LGTM
@IshwarKanse can you also take a look?

IshwarKanse (commented):
## Code Review — AI-assisted (CodeRabbit + Claude)

Overall the PR's intent is sound and the core changes are correct. A few issues worth addressing:

### Warning

**W1 — Test asserts ordering-dependent error string** (`cluster_test.go:92`)

```go
require.EqualError(t, err, `found multiple operator deployments for app.kubernetes.io/name="tempo-operator": ns-a/tempo-operator-a, ns-b/tempo-operator-b`)
```

The fake client's List ordering is an implementation detail, and a real Kubernetes API server returns items in unspecified order. If the fake client's internal ordering changes, this test will fail non-deterministically.

**Fix:** Sort the `deployments` slice before `strings.Join()` — this also makes production error output deterministic:

```go
sort.Strings(deployments)
return appsv1.Deployment{}, fmt.Errorf("found multiple operator deployments for app.kubernetes.io/name=%q: %s", c.config.OperatorName, strings.Join(deployments, ", "))
```

---

**W2 — `getOperatorNamespace()` mutates shared config, creating a hidden interaction with `getOperatorDeployment()`** (`cluster.go`)

`getOperatorNamespace()` caches the discovered namespace back into `c.config.OperatorNamespace`. After this PR, `getOperatorDeployment()` conditions its namespace filter on that same field. This means:

- The first call to `getOperatorDeployment()` (via `GetOperatorLogs`) runs **without** a namespace filter.
- A subsequent call after `getOperatorNamespace()` has populated the field runs **with** a filter.

Mutating shared config from within a discovery function makes call-order reasoning difficult. Consider caching the discovered namespace in a dedicated `cluster` struct field rather than in `c.config`.

---

### Info

**I1 — Missing test for the zero-deployments error path** (`cluster_test.go`)

Three tests are added, but the updated error message `"operator deployment with app.kubernetes.io/name=%q not found"` has no test. A test with an empty fake store asserting on that error would complete coverage of the PR's stated goals.

---

**I2 — Pod list in `GetOperatorLogs` is not namespace-scoped** (`cluster.go:103-106`, pre-existing)

The pod list query relies solely on the label selector. A pod in another namespace with the same labels could be returned. Adding `client.InNamespace(deployment.Namespace)` alongside the label selector would be the defensive fix.

IshwarKanse (commented):
## Code Review — AI-assisted (CodeRabbit + Claude)

Overall the PR's intent is sound and the core changes are correct. A few issues worth addressing:

### Warning

**W1 — Test asserts ordering-dependent error string** (`cluster_test.go:92`)

The fake client's List ordering is an implementation detail — a real Kubernetes API server returns items in unspecified order. If the fake client's internal ordering changes, this test will fail non-deterministically.

Fix: sort the `deployments` slice before `strings.Join()`:

```go
sort.Strings(deployments)
return appsv1.Deployment{}, fmt.Errorf("found multiple operator deployments for app.kubernetes.io/name=%q: %s", c.config.OperatorName, strings.Join(deployments, ", "))
```

**W2 — `getOperatorNamespace()` mutates shared config, creating a hidden interaction with `getOperatorDeployment()`** (`cluster.go`)

`getOperatorNamespace()` caches the discovered namespace back into `c.config.OperatorNamespace`. After this PR, `getOperatorDeployment()` conditions its namespace filter on that same field, creating an implicit call-order dependency. Consider caching the discovered namespace in a dedicated `cluster` struct field rather than in `c.config`.

### Info

**I1 — Missing test for the zero-deployments error path** (`cluster_test.go`)

The updated error message `"operator deployment with app.kubernetes.io/name=%q not found"` has no corresponding test. A test with an empty fake store asserting on that error would complete coverage of the PR's stated goals.

**I2 — Pod list in `GetOperatorLogs` is not namespace-scoped** (`cluster.go:103-106`, pre-existing)

The pod list query relies solely on the label selector with no `InNamespace` filter. Adding `client.InNamespace(deployment.Namespace)` would prevent pods in other namespaces with the same labels from matching.

PR #1532, opened 2026-07-01
[chore] add deploy skill

andreasgerstmayr:
A simple skill/script to (re)deploy the operator on Kubernetes, KinD or OpenShift.


.claude/skills/deploy/deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+    log "Detected platform: $PLATFORM"
+else
+    log "Using platform: $PLATFORM"
+fi
+
+case "$PLATFORM" in
+    kind|kubernetes|openshift)
+        log "Installing cert-manager"
+        make cert-manager
```
rubenvp8510, line 25: in which platform we don't need to install cert manager? I see all included here :D

rubenvp8510 (approved):

.claude/skills/deploy/deploy.sh
andreasgerstmayr: In the `openshift-olm` platform. Like openshift, but uses OLM to deploy the operator.
It's slower to deploy (needs to build and push bundle and build a FBC), so the default is the `openshift` platform.

PR #1543, opened 2026-07-16
chore: consolidate telemetrygen versions

andreasgerstmayr:


andreasgerstmayr:
will use telemetrygen from test-utils

PR #1550, opened 2026-07-23
feat(tempostack): default per-component replicas from size for high availability

rubenvp8510:
## Motivation

The TempoStack t-shirt sizes (`spec.size`) are documented as providing "HA support" for every non-demo size, but today only the **ingester** is actually made highly available — its replica count is derived from the replication factor quorum (`floor(RF/2)+1`). Every other component (distributor, querier, query-frontend, compactor, gateway, metrics-generator) defaulted to a **single replica regardless of size**, leaving the ingest and query paths as single points of failure.

## Changes

Non-demo sizes now default each component to a size-appropriate replica count so a single pod/node failure cannot break the ingest or query path. At least 2 replicas of every component, scaling the throughput-bound components (distributor, querier, metrics-generator) at the larger sizes:

| component        | pico | extra-small | small | medium |
|------------------|------|-------------|-------|--------|
| distributor      | 2    | 2           | 3     | 4      |
| querier          | 2    | 2           | 3     | 5      |
| query-frontend   | 2    | 2           | 2     | 2      |
| compactor        | 2    | 2           | 2     | 3      |
| gateway          | 2    | 2           | 2     | 2      |
| metrics-generator| 2    | 2           | 3     | 3      |

- The **ingester** is intentionally not in this table — it stays quorum-derived so the replication factor remains its single source of truth.
- **`1x.demo`** is unchanged and runs a single replica of every component (no HA).
- Explicit per-component `replicas` overrides always take precedence over the size defaults.
- The change is additive and backwards compatible: behavior only changes when `spec.size` is set and the user has not overridden a component's replicas.

### Implementation
- `internal/manifests/manifestutils/size.go`: new `ComponentReplicas` struct, `replicaProfiles` map, and `GetReplicaProfile` / `ReplicasForComponent` accessors (mirrors the existing `replicationFactors` pattern).
- `internal/webhooks/tempostack_webhook.go`: `Default()` now resolves per-component replicas from the size; ingester quorum logic untouched.
- Updated `spec.size` godoc, added a changelog fragment, and regenerated CRD/bundle/docs via `make reset`.

Note: in size mode, per-pod resource requests come from the fixed size profile and are **not** divided by the replica count, so each replica gets the full profile.

## Tests
- Unit tests for the new accessors (`size_test.go`) and webhook defaulting incl. scaled counts, override precedence, and no-size behavior (`tempostack_webhook_test.go`).
- Added `replicas` assertions to the e2e t-shirt-size assert files (including `1x.demo` = 1 to confirm HA is not applied, and the override test to confirm overrides win).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/webhooks/tempostack_webhook.go
```diff
@@ -148,22 +158,34 @@ func (d *Defaulter) Default(ctx context.Context, obj runtime.Object) error {
...
+		r.Spec.Template.QueryFrontend.Replicas = replicasForComponent(r.Spec.Size, manifestutils.QueryFrontendComponentName, defaultComponentReplicas)
 	}
 
 	if r.Spec.Template.Gateway.Replicas == nil {
-		r.Spec.Template.Gateway.Replicas = defaultComponentReplicas
+		r.Spec.Template.Gateway.Replicas = replicasForComponent(r.Spec.Size, manifestutils.GatewayComponentName, defaultComponentReplicas)
+	}
+
+	// The metrics-generator is only deployed when enabled; unlike the other
+	// components it is not defaulted to a single replica when no size is set, so
```
andreasgerstmayr, line 176: If `.Replicas` is unset, it defaults to `1`, right? I'd follow the same pattern for metrics generator as for the other components here.

andreasgerstmayr: nevermind, let's merge as-is

andreasgerstmayr (approved):

andreasgerstmayr:
> Merging is blocked
> Commits must have verified signatures.

You need to sign the git commit, otherwise we can't merge.

PR #1567, opened 2026-07-29
fix: reduce number of reconciles

andreasgerstmayr:
Reduce the number of reconciliations by filtering events on owned resources.

How these predicates differ from the default controller-runtime behavior (which triggers a reconcile on every create, update, delete, and generic event):

**createOrUpdateOnlyPred** (used on TempoStack/TempoMonolithic itself):

- Create: allowed (same as default)
- Update: only if `.metadata.generation` changed (spec change), annotations
  changed, or the object has a deletion timestamp — ignores status-only updates
  and other metadata changes like resourceVersion bumps
- Delete: suppressed (default would trigger)
- Generic: suppressed

**updateOrDeleteOnlyPred** (used on owned resources like ConfigMap, Service, etc.):

- Create: suppressed — the operator itself creates these objects, so there's no
  need to reconcile again when the create event fires back (the reconcile that
  created them is already done)
- Update: only on `.metadata.generation` changes (spec changes) — ignores status
  updates and label/annotation-only changes
- Delete: allowed only when the deletion is confirmed by the API server (not
  DeleteStateUnknown)
- Generic: suppressed

**updateOrDeleteWithStatusPred** (used on StatefulSet and Deployment):

- Create: suppressed — same rationale as above, the operator just created them
- Update: on `.metadata.generation` changes (spec changes) **or** when the
  status differs — this is needed so the operator can update the CR's status
  conditions based on rollout progress (e.g. ready replicas)
- Delete: allowed only when the deletion is confirmed by the API server (not
  DeleteStateUnknown)
- Generic: suppressed

**createUpdateOrDeletePred** (used on watched Secrets like storage secrets):

- Create: allowed — the operator needs to detect when a storage secret is created
- Update: on any `.metadata.resourceVersion` change — broader than
  generation-only, since Secrets don't increment generation on data changes
- Delete: allowed (same as default)
- Generic: suppressed

### DIFF
--- a/.chloggen/reduce-reconciles.yaml
+++ b/.chloggen/reduce-reconciles.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: operator
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Reduce the number of reconciliations by filtering events on owned resources
+
+# One or more tracking issues related to the change
+issues: [1567]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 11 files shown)


frzifus (approved):
N1, LGTM

PR #1576, opened 2026-08-04
Create a PodDisruptionBudget for each TempoStack component

rubenvp8510:
## Description

Creates a `PodDisruptionBudget` with `maxUnavailable: 1` for the distributor, ingester, querier, query-frontend, gateway and metrics-generator components of a `TempoStack`.

This ensures that voluntary disruptions, such as node drains during cluster upgrades, can only take down a single replica of a component at a time, keeping the rest of the component available.

Following the [Grafana Loki Operator](https://github.com/grafana/loki/tree/main/operator/internal/manifests), no budget is created for the compactor, and the same `maxUnavailable: 1` default is used.

## Notes

* The gateway and metrics-generator budgets are only created when those components are enabled in the CR. They are listed in `findObjectsOwnedByTempoOperator`, so they get pruned when the component is disabled.
* The reconciler now watches `PodDisruptionBudget` (`Owns`), and the `policy` API group was added to the operator RBAC (role and both bundle CSVs).
* `TempoMonolithic` is not affected by this change.

## Testing

* Unit test for the new `manifestutils.NewPodDisruptionBudget` helper.
* Object count assertions updated in `cmd/generate` and `internal/manifests`.
* The `e2e/compatibility` test asserts the four budgets created for that CR.

### DIFF
--- a/.chloggen/tempostack-pod-disruption-budgets.yaml
+++ b/.chloggen/tempostack-pod-disruption-budgets.yaml
@@ -1,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. tempostack, tempomonolithic, github action)
+component: tempostack
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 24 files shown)


frzifus (approved):

andreasgerstmayr (approved):
nice! :100:

PR #1578, opened 2026-08-04
Support zone-aware replication of a TempoStack

rubenvp8510:
## Description

Adds zone-aware replication to `TempoStack`, following the [Grafana Loki Operator](https://github.com/grafana/loki/blob/main/operator/internal/manifests/node_placement.go) in behaviour.

```yaml
spec:
  replicationFactor: 3
  replicationZones:
  - maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
```

Every component gets one `topologySpreadConstraint` per zone (`whenUnsatisfiable: DoNotSchedule`, selected by the component and instance labels), so the scheduler distributes its pods evenly across the topology domains. The ingester ring is configured with `zone_awareness_enabled`, so spans are replicated to ingesters in different zones and the stack survives the loss of a zone.

## How the availability zone reaches Tempo

A pod cannot read the topology labels of its node: the downward API only exposes `spec.nodeName`, not the node's labels. Same as in the Loki Operator:

1. The operator labels the zone-aware pods with `tempo.grafana.com/zone-aware` and annotates them with the configured topology keys.
2. A new controller (`tempostack-zoneaware-pod`) watches those pods, looks up the node they were scheduled on, joins the values of the configured topology labels and patches them onto the pod as `tempo.grafana.com/availability-zone`.
3. An init container blocks until that annotation is populated, exposed to it as a file via a downward-API volume.
4. The annotation is passed to the container as `INSTANCE_AVAILABILITY_ZONE` and lands in `ingester.lifecycler.availability_zone` — the containers already run with `-config.expand-env=true`.

The gateway is spread across the zones as well, but does not join the hash ring, so it gets neither the init container nor the env var. This mirrors the Loki Operator.

## Replication factor

`spec.replicationFactor` is unchanged and stays the single place to configure the replication factor. The Loki Operator nests both settings under `spec.replication` and deprecates its top-level field; that is not done here, since duplicating `spec.replicationFactor` as `spec.replication.factor` would mean deprecating a released field, which is a separate API decision and not required by this feature. The zones are therefore a flat `spec.replicationZones` next to it.

Make sure `spec.replicationFactor` is less than or equal to the number of available zones. As in the Loki Operator this is documented but not validated, since the operator does not know how many distinct values of the topology key exist in the cluster.

## Notes for reviewers

* **Compactor:** it is not part of `configureReplication` in the Loki Operator, so it is left out here too. Worth a second opinion, since the Loki compactor is a singleton while `spec.template.compactor` is scalable in TempoStack.
* **Deployment mutation:** `mutateDeployment` merged the pod template with `mergo.WithOverride`, which never overwrites an existing value with an empty one, so removing `spec.replicationZones` again would leave the spread constraints and the init container on existing Deployments and wedge scheduling. It now uses `mergeWithOverrideEmptyValue`, like `mutateStatefulSet` has since #880.
* **Init container hardening:** unlike the Loki Operator, the `az-annotation-check` init container gets `TempoContainerSecurityContext()` and the resource requirements of the Tempo container, otherwise it is rejected under the `restricted` pod security standard and by namespaces with a resource quota. Reusing the Tempo container's requirements does not increase the effective resource requests of the pod, since init containers run before the regular containers.
* **Gateway labels:** the gateway Deployment passed one label map to its `ObjectMeta`, its `spec.selector` and its pod template. `ConfigureReplication` adds the zone-awareness label to the pod template in place, which would have leaked it into the immutable selector, so the pod template now gets a copy. The other components are unaffected, they build their pod labels with `k8slabels.Merge`.
* **`maxSkew`:** bounded to `math.MaxInt32` in the CRD, since the field is a Go `int` that is narrowed to the `int32` of the topology spread constraint.
* **No e2e test:** exercising this needs a cluster whose nodes carry distinct topology labels. The unit tests cover the manifests, the rendered Tempo configuration and the new controller.

## Testing

* `internal/manifests/manifestutils/node_placement_test.go` — spread constraints, labels/annotations, init container (including its security context and resources), env var, volume, and the gateway exception.
* `internal/manifests/config/zoneawareness_test.go` — the rendered `ingester.lifecycler` section with and without zones.
* `internal/controller/tempo/tempostack_zone_labeling_controller_test.go` — annotating from single and multiple topology keys, unscheduled pods, missing node labels, missing annotation, deleted pods.
* `internal/webhooks/tempostack_webhook_test.go` — duplicate topology key validation.
* `internal/manifests/mutate_test.go` — removing zone awareness from an existing Deployment clears the spread constraints and the init container.
* `internal/manifests/gateway/gateway_test.go` — enabling zone awareness does not change the immutable selector of the gateway Deployment.

`make ensure-generate-is-noop` passes; the CRDs, bundles, RBAC and API docs are regenerated.


.golangci.yaml
```diff
@@ -25,6 +25,11 @@ linters:
       - linters:
           - staticcheck
         text: "(QF1001|QF1003|QF1008)"
+      # spec.replicationFactor is deprecated in favour of spec.replication.factor, but the operator
+      # still has to read and default it to stay backwards compatible.
+      - linters:
+          - staticcheck
+        text: "SA1019: .*ReplicationFactor is deprecated"
```
andreasgerstmayr: Can you add `// nolint` annotation in the code instead?

andreasgerstmayr (commented):
Overall looks good to me. As it's not straight forward to understand the interactions, could you add small markdown doc explaining how this feature works? Maybe create a new folder `docs/designdocs`?

By reading the PR, I think
> `TempoStackZoneAwarePodReconciler` reconciles Pods, and sets the `tempo.grafana.com/availability-zone` annotation based on the labels of the node where the pod was scheduled.
>
> Each pod gets an init container, which waits until `tempo.grafana.com/availability-zone` annotation is set on the pod. This works with the file volume mount, because using the env wouldn't work because env is only set at startup of the init container and not refreshed later. After the init container, the env var is used in the tempo config file.
>
> And all this is required because we can't set the availability zone in advance in TempoStack reconciler, because at that time we don't know the node yet where the pod(s) will be scheduled.

andreasgerstmayr (approved):
great feature! :100:

PR #1583, opened 2026-08-04
[chore] increase number of parallel chainsaw e2e tests

andreasgerstmayr:


andreasgerstmayr:
Unfortunately it didn't bring a significant decrease in test duration.

PR #1601, opened 2026-08-09
[chore] refactor query RBAC e2e tests, test /api/v2/traces API

andreasgerstmayr:
- Remove redundant gRPC ingestion testing (query path is the same regardless of protocol)
- Collapse trace generation and verification into fewer steps
- Rewrite verify scripts with helper functions and proper if/else
- Replace /api/v2/traces 403 check with actual RBAC filtering assertions (this API is now supported in gateway)


frzifus:
Checking later!

rubenvp8510 (approved):

PR #1606, opened 2026-08-10
Add option to disable gateway's ingress in OpenShift

### LINKED ISSUE #1416: Ingress of Tempo Gateway cannot be disabled in OpenShift cluster
We use the Tempo Operator (tempo-operator.v0.20.0-1) in an OpenShift cluster (OpenShift version 4.20.14). In our setup, we want to use the Tempo Gateway; however, we do not need the Route that gets created for the Gateway per default, and want to disable it for security reasons. So we set the respective section of our TempoStack manifest as follows, in accordance with the output of `kubectl explain tempostack.spec.template.gateway.ingress.type`:

```yaml
spec:
    template:
      gateway:
        enabled: true
        ingress:
          type: ""
```
However, when we apply our manifest, the value of `type` gets set to `"route"` in the cluster. It looks to me like the following part of `internal/webhooks/tempostack_webhook.go` may be responsible for our problem:

```go
    // if tenant mode is Openshift, ingress type should be route by default.
	if r.Spec.Tenants != nil && r.Spec.Tenants.Mode == v1alpha1.ModeOpenShift && r.Spec.Template.Gateway.Ingress.Type == "" {
		r.Spec.Template.Gateway.Ingress.Type = v1alpha1.IngressTypeRoute
	}
```

blanser-cc:
Resolves #1416 

## Description

This adds `none` as a value for `tempostasck.spec.template.gateway.ingress.type`. This serves to make it possible to disable the creation of an Ingress/Route for Tempo's Gateway even if the tenant mode is `openshift`.

In OpenShift clusters, if `tempostasck.spec.template.gateway.ingress.type` is not set (or set to the default value `""`), the Tempo operator will still create a Route for the Gateway, which is a useful default behavior in many cases. However, if no access to the Gateway is required via Route/Ingress, for security reasons it should be possible to disable the creation of the Route.

## Summary of changes

- Added `"none"` to `IngressType` enum constants and kubebuilder validation markers in `api/v1alpha1`.
- Implemented `(i IngressType) IsEnabled() bool` helper method to simplify reconciler checks.
- Regenerated CRDs and OLM bundle manifests (`make manifests bundle`).
- Added unit tests for `IngressType.IsEnabled()`.
- Added unit test cases for `TempoStackWebhook.Default()` when tenant mode is `openshift` and ingress type is either `""` or `"none"`.

## How to test

1. Deploy the operator locally or on a test cluster.
2. Apply a `TempoStack` CR with `spec.template.gateway.ingress.type: "none"`.
3. No `Ingress` or `Route` resource should be created for the `TempoStack`.
4. In contrast, for a `TempoStack` CR with `spec.template.gateway.ingress.type: ""`, a `Route` should still be created.

### DIFF
--- a/.chloggen/allow_to_disable_ingress_in_openshift.yaml
+++ b/.chloggen/allow_to_disable_ingress_in_openshift.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: tempostack
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make it possible to disable the gateway's Ingress when tenant mode is OpenShift
+
+# One or more tracking issues related to the change
+issues: [1416]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/api/tempo/v1alpha1/tempostack_types.go
+++ b/api/tempo/v1alpha1/tempostack_types.go
@@ -919,7 +919,7 @@ type JaegerQueryMonitor struct {
 // IngressSpec defines Jaeger Query Ingress options.
 type IngressSpec struct {
 	// Type defines the type of Ingress for the Jaeger Query UI.
-	// Currently ingress, route and none are supported.
+	// Supported values: ingress, route, none
 	//
 	// +optional
 	// +kubebuilder:validation:Optional
--- a/api/tempo/v1alpha1/ingress_types.go
+++ b/api/tempo/v1alpha1/ingress_types.go
@@ -2,19 +2,32 @@ package v1alpha1
 
 type (
 	// IngressType represents how a service should be exposed (ingress vs route).
-	// +kubebuilder:validation:Enum=ingress;route;""
+	// +kubebuilder:validation:Enum=ingress;route;none;""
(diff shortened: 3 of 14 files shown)


.chloggen/allow_to_disable_ingress_in_openshift.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. operator, github action)
+component: operator
```
andreasgerstmayr:
```suggestion
component: tempostack
```

api/tempo/v1alpha1/tempostack_types.go
```diff
@@ -888,7 +888,7 @@ type JaegerQueryMonitor struct {
 // IngressSpec defines Jaeger Query Ingress options.
 type IngressSpec struct {
 	// Type defines the type of Ingress for the Jaeger Query UI.
-	// Currently ingress, route and none are supported.
+	// Currently ingress, route, none and empty string are supported.
```
andreasgerstmayr:
```suggestion
	// Supported values: ingress, route, none
```

andreasgerstmayr (dismissed):

andreasgerstmayr (dismissed):

andreasgerstmayr (approved):
Thanks for the PR! :100:

andreasgerstmayr:
Unfortunately I can't merge it because some commits are missing a signature: https://github.com/grafana/tempo-operator/pull/1606#issuecomment-5239006245
Can you squash all commits into one, and sign it?

andreasgerstmayr:
Thanks! :)

